Issue #15, opened 2015-04-20
Kick off the design process for a proper front-end


gbinal:
- [x] Create basic front end layout and structure
- [x] Gather user data and information directed at the product
- [x] Write archetypes for our users (e.g. CIO)
- [x] Start developing various wireframe versions for the team to iterate over

PR #30, opened 2015-04-21
Initial commit for basic front end architecture and standard repo needs.

juliaelman:
This pull request includes a very simple front end architecture using the following tools:

- node/npm
- gulp
- bower

I added some documentation on how to get setup in your local environment and placed those in the README.md. Some other files to note:

- CONTRIBUTING.md
- LICENSE.md

Referencing #15, as this helps contribute to the completion of that issue.


.gitignore
```diff
@@ -0,0 +1,10 @@
+# Logs
+logs
+*.log
+
+# Dependency directories
+node_modules/
+bower_components/
+
+# Mac specific
+.DS_Store
```
msecret: Have you thought about what you want to do for compiled files? Some repos ignore compiled files, build then on the CI server, like travis CI, and then somehow get them to the actual server. This is more setup but better separation. It's easier to just include the compiled files in the repo but be aware that you may get more merge conflicts because of them.

scss/main.scss
```diff
@@ -0,0 +1,8 @@
+@charset "UTF-8";
+
+@import "../bower_components/bourbon/app/assets/stylesheets/bourbon";
+@import "../bower_components/neat/app/assets/stylesheets/neat";
+
+.container {
+	@include outer-container;
```
msecret: I think most projects are using 2 spaces for indents in css files. If you feel strongly about this though, you can comment on it on the CSS styleguide draft: https://docs.google.com/a/gsa.gov/document/d/1t6Rjslq9I20reFF3xfW3FVeV2NWFaDKauDj_LQJC5mc/edit.

.gitignore
juliaelman: Hrmm... I have not added compiled files before to the repo. Am assuming that we'd run a `bower install` and `npm install` when the CI is running, if we do so desire. It would be good to check that those commands work properly rather than test the actual compiled files themselves.

scss/main.scss
juliaelman: :+1: good catch and will update shortly!

msecret:
In terms of directory structure, where do you plan to put compiled JS when you get to that point?

One example of doing this would be:
```
src/js
src/css
public/js
public/css
```

Feel free to come up with your own structure though.

juliaelman:
@msecret yup! I am thinking we'll most likely add an `assets` folder and place all static folders (scss, css, js, images, fonts, etc.) inside of there. For now, just wanted a super basic structure put together for the team to start hacking away on.

There are also some nice compression tools we can use with gulp to minify JS files. Once we get started on building out that structure more, I'll be adding that tooling in.

Thanks again for the taking the time to review! :fireworks:

konklone:
I resolved merge conflicts with the changes I made in #36, and reviewed. I have some more changes I'd like to make, that'll come in a separate PR.

PR #32, opened 2015-04-22
Frontend jquery

vzvenyach:


juliaelman:
Looks great! :+1: There are a few things I noted directly in the code and also have a few more notes here:

1. Looks like there is a merge conflict. Happy to help with that if you'd like me to, but want to make sure to resolve that before merging in.
2. What do you think about creating a `data` folder and placing the `json` files in there instead? Sort of like how the hub does it: https://github.com/18F/hub/tree/master/_data

vzvenyach:
@juliaelman makes sense about putting it in `vendor` vs. `bower`. Will replace this with a different approach...

juliaelman:
Closing this in favor of #33 (which was merged into `basic-frontend`).

PR #36, opened 2015-04-25
Clean up docs, open source licensing

konklone:
The `.md` docs we had are all now memorialized in other places. This also adds 18F's [standard open source licensing stuff](https://github.com/18F/open-source-policy).


konklone:
As a non-code, organizational PR, I'm going to self-merge so I can integrate the outstanding branches more cleanly.

PR #47, opened 2015-04-28
Production webhook in place

konklone:
This adds a tiny system for a webhook that hits our web server's nginx process, which is proxied to a local hookshot instance which runs a `git pull` on the directory where our static site is.

The `hookshot.js` file was added to `master` instead of `production` by mistake in https://github.com/18F/dotgov-dashboard/commit/4a14218cf57fc3e120dd144a26f1587e525e3960. 9_9 The rest is here.

It's not an ideal system, but it's the one used by 18f.gsa.gov and https.cio.gov, and very very similar to the one used by pages.18f.gov. I can imagine us moving to a CloudFoundry-based system like we're setting up for staging, after we evaluate the tradeoffs there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,3 +10,5 @@ deploy:
 script:
 - npm install
 - bower install
+
+sudo: required
--- a/deploy/fabfile.py
+++ b/deploy/fabfile.py
@@ -1,0 +1,43 @@
+import time
+from fabric.api import run, execute, env, cd
+
+"""
+Manage auto-deploy webhooks remotely.
+
+Production hook:
+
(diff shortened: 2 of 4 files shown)


konklone:
To be clear, this makes it so that every time something is pushed to the `production` branch, the production site updates to be its contents. This PR is syncing the work I did on the `production` branch back to `master`, and the PR's branch should **not** be deleted afterwards. From now on, PRs should be done to `master` for all development work, and production deploys can be done via PR from `master` to `production`.

konklone:
Self-merging now that staging and production are both confirmed as auto-deploying, so we can move on.

PR #49, opened 2015-04-29
Deploying DataTables front-end to production

konklone:
It's ugly! But it functions.


juliaelman:
Looks good! :+1: Added `fast-cv` to `package.json` as a dev requirement.

PR #63, opened 2015-05-04
Create dataHandling.py

jtexnl:
This script serves two functions: first it compares data from the site-inspector scan (here 'inspect.csv') to the data from the registrar ('marchDom.csv') and then it reshapes the data into an output that can be more easily fed into a D3 visualization ('output.csv'). I would usually do this kind of script in R, so it's possible that the code can be cleaned up.

### DIFF
--- a/dataHandling.py
+++ b/dataHandling.py
@@ -1,0 +1,46 @@
+import pandas as pd
+
+path="/Users/JohnTindel/Documents/Work/Data Library/"
+
+scan = pd.read_csv(path+"inspect.csv")
+marchDoms = pd.read_csv(path+"marchDom.csv")
+
+liveComp = scan[['Domain', 'Live']]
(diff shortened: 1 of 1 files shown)


konklone:
I hate closing a PR, but we talked about two things that need to be done:

* De-hardcoding paths, etc. (`/Users/JohnTindel`)
* Splitting up the responsibilities: transforming the data is an ongoing need, evaluating/comparing whether sites are live is a one-off need.

We're at the point where we do need data transformation scripts, and likely in Python, that will take the output of domain-scan and produce some of the aggregations we need. I (and I think @juliaelman) are leaning towards making those scripts in this repo. Let's come back at this problem again, this week, and make a Python script that uses STDIN/STDOUT to produce the aggregated JSON we need to render data client-side in the app.

PR #67, opened 2015-05-05
Production- Change footer text on live site

jtexnl:
Could we make the placeholder text at the top of the pulse page a little bit more official sounding in case someone who comes across the page is not technical? There has been some concern that the pre-alpha language might be confusing to non-technical folks. I know we're not publicizing anything yet, so no outside parties should get onto the site in theory, but some folks at OGP are uneasy about the language as it stands nonetheless. I'd suggest "This website is in a very early stage of development. No information or content found here should be considered definitive."


konklone:
This PR is in the wrong direction, and has no content. See https://github.com/18F/dotgov-dashboard/issues/66#issuecomment-99122348 for our preferred workflow -- very happy to come and help walk through it.

Issue #79, opened 2015-05-07
Add Jekyll to repo setup and README.md


juliaelman:
After coming to a more well defined version of the wireframes, @konklone agreed that moving forward with implementing Jekyll is something that makes the most sense. Here are some of the tasks to accomplish to complete this work:

- [x] Add `Gemfile` to repo (or whatever would work best) to include Jekyll into the setup
- [x] Add any additional installation instructions to `README.md` to guide any new or existing contributors on how to use this new setup

juliaelman:
@konklone adding you as the implementer for this issue, but let me know if you need anything else from me!

konklone:
The infrastructure work was done in #80, and I've since re-configured our Travis, Cloud Foundry, and webhook setups to restore auto-deploy to staging and production. All done!

PR #80, opened 2015-05-07
Jekyllization

konklone:
This moves the site over to Jekyll, in preparation for an anticipated multi-page app.

It removes `bower`, `npm`, and `gulp` in favor of installing `sass`, `neat`, and `bourbon` through gems in the Gemfile that we now already need to have for Jekyll.

The main `index.html` file is now split up into a template. There's a super-basic `_config.yml` that uses `redcarpet` as the markdown processor, and specifies `rouge` for syntax highlighting. Even though we don't have any specific plans for syntax highlighting, I'd like to be explicit about never requiring pygments, so that a subtle Python 2 dependency is not ever present.

I also replaced the `gulpfile.js` with a `Makefile` in the style of [what we did for analytics.usa.gov](https://github.com/GSA/analytics.usa.gov/blob/master/Makefile), though I used `bundle exec` for the `sass` command.

These were some opinionated choices! I did them to make them more consistent with our other Jekyll sites, and to try to minimize the number of build systems and system dependencies we had in place (since we added one with gem/Bundler/Ruby). We can always add some back in if the team wants.


konklone:
Oh, a note -- this breaks the auto-deploy process for our staging and production sites. Restoring those is in my next PR!

README.md
```diff
@@ -2,48 +2,38 @@
...
+bundle install
 ```
 
-* Now you should be able to run `gulp` and open the `index.html` file in your favorite browser.
+* Now you can run the app:
 
 ```bash
-gulp
+make run
 ```
```
juliaelman: It might be good to add in something in here on the `make` commands provided and what they do. For example, if someone is working on the styles and needs to compile the sass. `make run` is not currently setup to compile and will not show the updated styles.

juliaelman:
Looks good! :+1: Quick note that we'll need to be sure to update our `.travis.yml` to not run the dependency script in place for the previous setup. This should also satisfy all requirement for #79! :rainbow:

PR #82, opened 2015-05-10
Add Google Webmaster Tools verification code

konklone:
This lets us see how Google sees our site, and offers a bunch of helpful analysis.

Access to a verified profile can be shared among multiple teammates, so we should only have to do this once.


juliaelman:
Looks good! :+1:

Issue #75, opened 2015-05-07
Homepage front end implementation


juliaelman:
Below is the PNG layout for the Pulse homepage. Please note that the numbers are not accurate in this sketch are are merely for positioning. Accurate figures will be part of this implementation and are listed as one of the tasks below:

- [x] Create initial stylesheets for overall look/feel for the site (basic level; Bourbon/NEAT)
- [x] Add top navbar with corresponding links to pages
- [x] Check-in with the design team about header/footer usage as seen on other 18F projects (e.g. openFOIA)
- [x] Add suggested header/footers per design direction to Sketch files + code
- [x] Collect aggregate data on the homepage for overall HTTPS and DAP stats
- [x] Add donut chart implementation of the data 

![homepage](https://cloud.githubusercontent.com/assets/955558/7573576/b3b61026-f7f4-11e4-964c-44d9abad511f.png)

konklone:
Work is ongoing in the `homepage` branch. The other wireframe tickets (#76, #77, #78) should not be worked on until this branch is merged in, as it contains most of the initial framing and layout. Sorry to be a blocker! I'll try to get the frame merged in first, before the homepage is fully done and this ticket can be closed, and release the block.

juliaelman:
@konklone I reviewed the PR, made some changes and merged it in. Thanks for knocking out the first three tasks on this issue :+1:

juliaelman:
Overall feedback from the design team was to place the current footer disclaimer at the top of the page, rather than the bottom.

konklone:
:+1:

konklone:
Closed with #90.

PR #83, opened 2015-05-11
Moving official US Govt Website tag to the top, refs #75.

juliaelman:


_includes/disclaimer.html
```diff
@@ -1,5 +1,12 @@
-<div class="disclaimer">
+<header class="site-header">
```
konklone: Why a `site-header` class here?

juliaelman: B/c `<header>` may exist on other pages we are building and needed to add some sort of differentiation to this one.

konklone: But those can be distinguished by hierarchy. My understanding of `<header>` as a semantically named tag in HTML5 is that it should not appear twice in the same level of hierarchy, which ensures it should never need a class name.

juliaelman: Agreed. But since the hierarchy on this is really `<body>` it might be nice to distinguish it specifically from other headers. This is similar to the way that openFEC did implemented it: https://github.com/18F/openFEC-web-app/blob/develop/templates/layouts/main.html#L63

Happy to move forward without it, but wanting to try to think ahead for future codings.

konklone: It's definitely not worth holding up a merge over. ;) So I'll merge it in. But using a class name seems against the point of having `<header>` in the first place (same with `<nav>` or `<footer>`). I think the appropriate selector here is `body > header`, which IIRC selects an immediate child (not a more distant descendant).

Issue #78, opened 2015-05-07
DAP stats page front end implementation


juliaelman:
Below are two PNG layouts for the Pulse DAP pages. A few tasks that need to be completed are:

- [x] Dynamically retrieving the DAP compliance stats into the page and placing into a dataTable
- [x] Copy at the top that explains how we came to this aggregate data and what DAP is
- [x]  Grouping the current data set by agency

![dap - filtered by domain](https://cloud.githubusercontent.com/assets/955558/7573598/dae0e3c4-f7f4-11e4-96dc-840f4ba59e55.png)

![dap - filtered by agency](https://cloud.githubusercontent.com/assets/955558/7573600/dc59c1da-f7f4-11e4-82cb-c05f30b70d8e.png)

juliaelman:
Making not of a comment made by @konklone from round 3 wireframes feedback:

"I personally think the definition and methodology behind the terms we grade agencies on will be necessary for MVP (though not necessary for Sprint 2 / these wireframes) in order for the site to be fair and helpful."

konklone:
Closed by #110.

PR #85, opened 2015-05-13
A start at an /analytics/ page

konklone:
This takes care of some (not all) of #78, as well as modularizing some (not all) of our D3 donut code.

screenshot from 2015-05-12 21 17 13

The code is harmonized a bit between the HTTPS and DAP JS includes, by making names consistent and wrapping each in an anonymous function call so that their variables and namespaces don't collide.

The width of the donut is now either read in from a `data-width` attribute on the `.https_chart` or `dap_chart` elements, *or* if one is not provided the code looks at the parent element of that chart and grabs it's calculated width using [`getComputedStyle()`](http://caniuse.com/#search=getComputedStyle).

In both cases, the height is now dynamically derived from the width, using the original hardcoded sizes, for a ratio of 1:1.2.

This re-embeds datatables, with the DAP code, and attempts a teensy bit of styling, like making the select box purple. However, the arrow is black, and to complete the styling will require setting `-webkit-appearance: none` on the `<select>` and then making an arrow *icon* image and setting that as the background-image. Happy to defer to @juliaelman on the best set of tradeoffs/work there.


konklone:
I'll add that I'm not a tremendous fan of how heavy datatables is for us, and I'd say I'm "in the market" for a lighter or more flexible one. I don't mind favoring a more D3-ish one that provides the internal modeling and hooks, while relying on the developer to do the wiring.

konklone:
Oh, more majorly -- this has not implemented the domains vs agencies nav, or the agencies table.

juliaelman:
@konklone this is a great start and like the implementation of modularity to the donut chart. :+1: Going to go ahead, merge in and will make any styling changes as we move forward.

As for dataTables, 100% agree that it's something we should consider for later phase implementations. would be good to connect with others in the frontend team for suggestions/best approaches too!

Issue #76, opened 2015-05-07
About page front end implementation


juliaelman:
Below is the PNG layout for the About page. A few things that we discussed today and that also need to get accomplished in partnership with this issue:

- [x] Complete layout implementation in template
- [x] Write copy for Frequently Asked Questions

![about](https://cloud.githubusercontent.com/assets/955558/7573588/c0ba0250-f7f4-11e4-8472-927fe2bd788c.png)

juliaelman:
@konklone opened #86 for styling implementation and assigning this over to you for copy.

juliaelman:
Initial copy has been written, but we still need to think about where to point folks to in order to to get better scores. Moving this into Sprint 3.

konklone:
@juliaelman Since we're tracking the score/implementation concerns in #113, is this good to close?

juliaelman:
@konklone +1 and closed.

PR #86, opened 2015-05-13
A start at an /https/ page

konklone:
This takes some of the templating from #85 and applies it to `/https/`. It includes a first pass at a description for the top of the HTTPS page.

screenshot from 2015-05-12 23 28 00

There's extensive data filtering/transformation here, to get to the nomenclature we discussed for our IA on HTTPS categorization. After writing this, I'm more convinced than ever we need to get a (Python-based) data pipeline going, in this repo, that takes the base dataset produced by `domain-scan` and produces the data files we need for direct rendering as a table. This is so that we can capture as much business logic as we can outside of front-end JS, where it's inevitably intermingled with display concerns. In the meantime, it's being done in the front-end.

It includes some sample HTTPS data. The next areas of work are clearly data pipelining, to make this less ad hoc and with 100% real data.


konklone:
screenshot from 2015-05-12 23 28 00

konklone:
Another thing I noticed - metadata about the domain's agency and branch need to be carried through the data pipeline, in a way they are currently not. There's no other good way to evaluate a domain's eligibility that way.

I'm currently filtering out domains which are not "Live" for HTTPS. DAP should filter out domains that are either not "Live", or are just redirectors, or are non-executive. This filtering should be applied to both the aggregate %, and the rows in the table.

It feels very silly to be parsing JSON that is transformed from a CSV which was transformed from JSON. `domain-scan` should produce JSON output as well, which this site's specific data pipeline can parse into direct-to-render JSON.

jtexnl:
For 508-compliance, I would give the link text color more contrast to the background. The words "entirely to HTTPS" are not distinct enough to the colorblind eye.

juliaelman:
Looks good :+1: 

@konklone all good points and completely makes sense to move towards a Python based solution for the data pipeline! :+1: Maybe we add that as part of a task for Sprint 3? 

@jtexnl agreed on the color of the link in the header. I was thinking of making it white with an underline, but that strays from the overall styling for links site wide. I'll incorporate a fix for that in the "About" page work I'll be doing to help complete #76.

PR #90, opened 2015-05-15
Real data for /https/ and /analytics/

konklone:
screenshot from 2015-05-15 11 26 58

This updates the rendered tables with our last known complete set of data, with a first pass at language for HTTPS that describes it all.

It also applies the correct filters to our data:

* DAP-eligible domains are only live, non-redirect, executive branch domains.
* HTTPS-eligible domains are only live domains.

A new `data.py` script takes the .gov domain list, and the result CSVs of domain-scan, and creates the necessary JSON files that the table uses. It attaches enough data for the JS to make the above eligibility determinations client-side.

The new % numbers for the donuts is hardcoded, but now correct, based on the other numbers.

We still have more proper data pipelining to go, but this shows our real numbers in our front-end buildout.


konklone:
screenshot from 2015-05-15 11 27 24

screenshot from 2015-05-15 11 27 30

PR #109, opened 2015-05-19
More data pipeline work

konklone:
This PR:

* Refactors the `data.py` data preparation script to be broken into smaller pieces that operate more clearly.
* Simplifies the HTTPS column values into more clear Yes/No values.
* Renames the HTTPS and DAP column headers to be more user friendly.
* Moves data filtering and transformation out of JavaScript and into Python.
* Renames and standardizes output of the CSV files that the donuts draw from.
* Generates those CSV files from the real calculated %'s, in Python.
* Removed hardcoded donut %'s and uses those specified in the CSV.

This creates much steadier groundwork for agency aggregation and other kids of data transformation and publishing.


konklone:
Oh, it also:

* Removes the CSV inputs from `domain-scan` from the versioned repository, but leaves the currently used domain list versioned, so it's clear what was used.
* Now takes the full .gov CSV as input (instead of just federal domains), and filters out non-federal domains while reading domains. (I've also re-added `18f.gov` by hand into the latest snapshot.)
* Re-adds the search box above data tables.

Most of this was invisible to the front-end. Here's a screenshot of one small page's worth of the HTTPS details, with the new column headers and field values:

screenshot from 2015-05-19 02 13 50

juliaelman:
Nice work! :+1:

Issue #106, opened 2015-05-19
Agency tabs and data grouping


konklone:
A "by agency" tab on the HTTPS and DAP landing pages that groups data by agency, rather than by domain.

This involves both front-end buildout work, and data pipelining work to create data grouped appropriately for this view.

juliaelman:
Both #77 and #78 are effected by this. Just noting here so that we can keep those in-sync with this issue.

konklone:
Closed by #110.

PR #110, opened 2015-05-19
Add agency tabs to HTTPS + Analytics pages

### LINKED ISSUE #77: HTTPS stats page front end implementation
Below are two PNG layouts for the Pulse HTTPS pages. A few tasks that need to be completed are:

- [x] Aggregate numbers for total HTTPS compliance to complete the donut chart
- [x] Copy at the top that explains how we came to this aggregate data
- [ ] Grouping the current data set into higher level categories as outlined in the wireframes
- [ ] Grouping the current data set by agency

https - filtered by domain

https - filtered by agency

juliaelman:
This pull request includes adding agency tabs, pages and static data for agency information to the HTTPS and Analytics data.

Fixes #77, #78, and #106.


konklone:
OK, I've made it real, with a few changes:

* I made the `border-radius` uniform on the % indicators, as it was becoming a flat line on the right when at 100%.
* I renamed the URLs and underlying directory structure in all sorts of places, to use directories (`/`) rather than filenames (`-`) to separate things. So it's `/analytics/domains/` and `/analytics/agencies/` for the page URL, `/assets/data/tables/analytics/domains.json` and `/assets/data/tables/analytics/agencies.json`, etc. I also rearranged the HTML templates and includes the same way.
* I updated `data.py` to aggregate data based on agency and output the data to the right place.
* I added a `utils.js` file that is currently loaded in everywhere, right now to provide a progress bar helper. If this turns out to be overkill, we can remove it. With our current production deployment (which supports SPDY), there's negligible overhead in doing it this way.

real

konklone:
(Oh I also adjusted some styling on the purple dropdown for number of items to see.)

One main issue I see with our current approach to agencies is that we're using the raw data that comes out of the .gov domain list. This is natural, and probably the only feasible way to do this right now -- but it also means we have agency distinctions that aren't used elsewhere. For example, the USPS and the USPS IG are separate. And all the agencies that start with "Department" get alphabetically moved into the D's and off the front page. This is particularly rough for the DAP view, whose whole first page of agencies has 0 support.

konklone:
I just updated the PR to default the By Agency tables to sort by "Number of Domains", from highest to least. This solves the alphabet issue, while floating the highest footprint agencies up to the top.

konklone:
I made one more under-the-hood change to convert the underlying cell data we store in JSON to be numbers, rather than display names. This allows for sorting measurement columns by quality, rather than by alphabet, and means we can keep more granular distinctions in scenarios, while showing the same display name for both cases if we want to.

juliaelman:
Nice work and like how the files are now organized! :rocket:

PR #117, opened 2015-05-20
Update production with agency data and other fixes

konklone:
This syncs the production site to the staging site. It includes a raft of design improvements, and additional agency-grouped data.


juliaelman:
:+1: looks good!

PR #119, opened 2015-05-21
Updating donut verbiage on homepage and linking up the donuts

### LINKED ISSUE #116: should donut charts be hyperlinked?
On the homepage, I find myself wanting to (and trying to) click on the donut charts (or at least the numbers in the middle of them) to get to /https and /analytics.  If it's not hard, I think that would be useful to folks.

juliaelman:
Fixes #116 and updating text per user feedback.


konklone:
Awesome, thanks! I just added a commit to make the mouse cursor a pointer when over the donuts, since they're clickable now. That seems minor enough not to hold up the merge.

PR #122, opened 2015-05-22
Removing purple theme and replacing with blue/green theming.

juliaelman:
Before:

before

After:

after


konklone:
Thanks, @juliaelman. I posted before/after screenshots in the ticket description above.

PR #123, opened 2015-05-22
Add SSL Labs grades

konklone:
This adds SSL Labs grades to the two HTTPS tabs.

* On the per-domain list, it shows the grade, and links over to the SSL Labs report for that domain. The link is styled to be underlined, and always opens a new tab.
* On the per-agency list, it groups the % of that agency's domains which have at least an `A-` grade.
* For domains not using HTTPS, it shows no grade at all. I also applied this logic to the Enforcement and HSTS columns, so now they're blank if the domain does not have HTTPS enabled at all.

Two screenshots below:

domains

agencies


juliaelman:
Looks good! :+1:

Issue #113, opened 2015-05-19
Call out 'How to Do Better!' information 


gbinal:
I think we should strongly call out in the headers of /https and /analytics the paths by which agencies can do better.  Perhaps:  

"Need help adopting HTTPS?  Go here!" [link to section of https.cio.gov that specifically offers help, needs creating I think.]

"Need help joining DAP?  Go here!" [link to http://www.digitalgov.gov/services/dap/]

juliaelman:
Just making note, this is one of the tasks outlined here: https://github.com/18F/dotgov-dashboard/issues/76#issue-73821162

konklone:
:+1: to all of this. Closing in favor of continuing to track in #76.

konklone:
Er, wait, we also said we'd split this out into a separate nav item and landing page. Which I think is smart. So let's track this here.

juliaelman:
Resolved by #125. There will most likely be some refining, but this generally covers the purpose of this issue.

PR #125, opened 2015-05-23
Guidance for each measured category

konklone:
A first pass at a `Guidance` tab on the HTTPS and DAP pages, that explains what's going on, and each evaluated field.

It could use some more work, and some more styling. @juliaelman, definitely open to tearing up the layout and approach.

DAP:

dap

HTTPS:

https

Contributes to #113.


juliaelman:
Looks good! :+1: Made a few changes to the page structure and styling.

PR #127, opened 2015-05-25
Move sort arrows from right to left of column header

### LINKED ISSUE #98: Sort buttons are placed in a confusing place.
On https://pulse.18f.gov/https/, the up/down arrows are so far to the right that on first glance, I keep thinking that they are tied to the column header to the right and not to the left.  So when I think I'm sorting by `HTTPS Behavior`, I actually sort by `HTTPS`.  

screen shot 2015-05-18 at 2 21 29 pm

konklone:
Moves the sort arrows from right to left. This expands leftwards padding on the cells by 20px, too. It's possible we may wish to adjust this styling later to only apply to `<th>` and not `<td>`.

Before:

screenshot from 2015-05-25 17 07 37

After:

screenshot from 2015-05-25 17 08 00

Fixes #98.


konklone:
This is also downstream of #125, which should be merged first.

juliaelman:
Switched up the coloring of the arrows to match the color scheme! Looks good.

Issue #95, opened 2015-05-15
Responsive fixes for various different layouts


juliaelman:
- [x] Determine screen widths for break points
- [x] Implement break points in stylesheets
- [x] Create fixes in various pages
- [ ] Test on various devices

konklone:
For the table itself, I think we're going to want to use the `Responsive` extension:

https://www.datatables.net/extensions/responsive

I've looked through its docs, and I think it'll be acceptable. Prioritizing other stuff until post-27th, but happy to field that part of this ticket after that.

juliaelman:
Current breakpoints are: 480px, 767px, 1024px and 1200px.

PR #129, opened 2015-05-25
Responsive tables

konklone:
This is a first pass at using DataTables' `Responsive` extension to make our tables tolerable in constrained viewports.

@juliaelman, do feel free to restyle all this stuff, including the green "more" button thing (which is done in CSS, not an image). We could also probably make it so that the agency % bars style differently in the expanded view than they do in the table view.

Again, I'm not thrilled with DataTables, but we don't seem to have a better viable alternative for this set of work at the moment, and this is the best framework we'll be able to work in for this timeframe.

**Note:** The "More details" button that appears when the viewport is constrained can *also* be made to be present all the time, by adding columns whose classes are `none` and will *only* be shown when 
the button is pressed. This isn't used here yet, but this lets us piggyback an expanded details view onto this foundation, which will be helpful.

This PR styles all 4 tables, picking different breakpoints for each. I also locked the `Domain` column to a fixed width of `210px`, and disabled whitespace wrapping to prevent `9-11commission.gov` and other hyphenated domain names from getting needlessly wrapped.

Screenshots of some tables in different sizes:

HTTPS domains on phone:

screenshot from 2015-05-25 18 03 05

HTTPS domains on tablet:

screenshot from 2015-05-25 18 03 19

HTTPS domains on desktop:

screenshot from 2015-05-25 18 03 29

HTTPS agencies on phone:

screenshot from 2015-05-25 18 19 43

HTTPS agencies on tablet:

screenshot from 2015-05-25 18 19 33

I didn't screenshot the DAP table, because it's far simpler, but I added responsive options there too.

Contributes to #95.

This is downstream of #127, because the arrow padding was something this built on. #127 should be merged first.


alex:
I'm not sure red/green are good colors for the expand/contract icons. They look too much like positive/negative incidcators.

konklone:
> I'm not sure red/green are good colors for the expand/contract icons. They look too much like positive/negative incidcators.

Totally agree. We could handle that here pre-merge, or post-merge.

juliaelman:
PR looks good! 

@konklone @alex changed the button color scheme to blue/red (see screenshot below). If this looks good to everyone, I'll go ahead and merge in:

screen shot 2015-05-26 at 6 58 45 am

PR #133, opened 2015-05-27
Deploy latest work to production

konklone:
This deploys the guidance tabs, responsive tables, and a host of smaller tweaks, to the production site from staging.

Let's merge this when we have the version on `master` that we want present during our CIOC presentation. This PR can stand open as we merge other PRs into `master`, and will be updated in place automatically. 

However, let's merge it no later than a half hour before that presentation, so that if there are technical hiccups during the deployment process, they have time to be resolved.


konklone:
Significantly, this will deploy SSL Labs grades to production.

PR #143, opened 2015-05-27
Cache-bust our JS and CSS

### LINKED ISSUE #92: CSS should use cache busting technique
e.g. referencing it as `<link rel="stylesheet" href="/assets/css/main.css?{hash}">`. Right now if the CSS changes, users' browsers may use out of date CSS. (This just happened to me :-))

konklone:
Cache-busts the CSS and JS files we author in this repository, at site build-time. JS and CSS will still be cached appropriately between site updates/builds.

Fixes #92.


juliaelman:
:+1: :shipit:

PR #144, opened 2015-05-27
Make DAP script include asynchronous

### LINKED ISSUE #128: Don't let table loading depend on analytics connections
Connections to `analytics.usa.gov` and `google-analytics.com` delay the table from rendering. This should be avoided if at all possible, especially for mobile contexts.

konklone:
GA is already using `async`, but DAP was not. This adds the `async` attribute to the DAP include.

This should be replicated across our other sites using the DAP snippet -- best to do it when the domain changes, and we already want to go around updating snippets one last time.

Fixes #128.


juliaelman:
:rocket:

PR #147, opened 2015-05-27
Revise HTTPS % methodology

konklone:
This now excludes domains which "downgrade" HTTPS (redirect from `https://` to `http://`) from the calculation of "uses HTTPS" that we make our big % out of.

It drops the % from 34% to 31%. However, it is a more honest assessment of whether a domain Uses HTTPS.

I'm still working on this, to sync the language and guidance on the table. Will update when ready for merge.

### DIFF
--- a/assets/data/https.csv
+++ b/assets/data/https.csv
@@ -1,3 +1,3 @@
 status,value
-active,34
-inactive,66
+active,31
+inactive,69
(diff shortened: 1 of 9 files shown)


konklone:
OK, I've updated the Guidance tab, renamed the columns to `Uses HTTPS` and `Enforces HTTPS`, and rejiggered the methodology so that the `Uses HTTPS` column takes downgrading into account.

This does cause the table to exclude HSTS and SSL Labs values for sites which technically have valid HTTPS configured, but which redirect down to HTTP. I think that's an acceptable UX tradeoff, and reduces confusion substantially, for the ~55 domains which downgrade to HTTP.

PR #149, opened 2015-05-27
Add an extended Details field for HTTPS domain list

konklone:
This adds a _first pass_ at a "Details" field for domains, which right now is just a sentence characterizing a domain's support for HTTPS. 

It says if a site is downgrading HTTPS, and if there are certificate chain issues it mentions that too. For domains with cert chain issues, it changes the value to `Yes*` with an asterisk, since we can now explain the asterisk.

The goal with this section is to expand it to explain more things about the other columns, and to offer a spot to link to specific issues (like how to fix a cert chain, or fix an HSTS header), as well as to link to the data behind our analysis. 

This is where we can provide domain-specific advice to program owners and CIOs. I'm adding in something, even though it should be expanded, so we have a clear place for advice to which we can point people viewing the site today.

It removes the media query from around the padding on the + button, and makes it appear on the HTTPS domains tab at all times. Example:

screenshot from 2015-05-27 13 07 46


juliaelman:
Looks good! Added some spacing and am a bit worried about the closeness of the plus sign and the link to the url. This will definitely need some more design thought around it and am opening a new issue to address at another time.

Issue #118, opened 2015-05-20
Ability to expand websites under a given agency?


alex:
It seems like it'd be useful to be able to expand websites under an agency, so you can see which sites have the features, and which do not.

juliaelman:
Thanks for the suggestion! Maybe we could think about a Phase One step and add an "Agency" column to the "By Domains" tab?

konklone:
I wonder if we can support permalinks to specifically filtered lists, like @gbinal said. If so, we could link from the agency list's "No. of Domains" amounts to the domain list filtered to that agency.

gbinal:
Yeah - I didn't appreciate it enough until now but def. realize that a major use case for both the by domain and by agency tabs will be "I want to see my agency's domains".  

I think something that gets us by will be pretty important with this.  Thinking about more.

konklone:
Moving to Sprint 3, and hopefully this is a simple addition after #101 is resolved.

gbinal:
Thinking about this just now and two 'easy' solutions' seem to me to be:  
* Add a column next to `by agency` (one way to possibly keep it narrow and tidy would be to use [the social media registry api](http://registry.usa.gov/agencies.json) to provide acronyms and have tooltips show the full agency name).  
* Just add the agency acronym in brackets after the domain, e.g. `18f.gov [GSA]`.  That way folks can use the quickfilter to find their agency.  

The 2nd idea isn't perfection but it might be easiest and would do to get us off the ground.

gbinal:
P.S. The api endpoint I linked in the comment above has some ability to set query parameters.  For more, check out the `/agencies` references on [this page](http://www.usa.gov/About/developer-resources/social-media-registry.shtml).

konklone:
Bear in mind we're using agency names that come straight out of the .gov registry data, some combination of GSA and Verisign data. So like, USPS and USPS OIG are listed separately. It's not standard, and I'm sure we'll get questions, but we also can't cleanly and automatically map our data onto standard acronyms as a result.

konklone:
I think starting from the Agencies tab and linking the number of domains, e.g. "16 domains", and having that bring the user to a filtered Domains tab to that agency, would probably be a simple place to start, but continue to be open to other ideas. =)

gbinal:
> I think starting from the Agencies tab and linking the number of domains, e.g. "16 domains", and having that bring the user to a filtered Domains tab to that agency, would probably be a simple place to start, but continue to be open to other ideas.

[...]

PR #156, opened 2015-05-28
Link number of agency domains to a filtered list of their domains

### LINKED ISSUE #101: enable permalinks
I'm guessing this isn't readily doable for launch, but it would be great if every page customization (number of entries shown, what page of results you're on, sorting) all changed the URL so that permalinks were available by default.

konklone:
This adds permalink support for search filters, and uses this feature to add support for filtering domains by agency.

image

Clicking the HHS domain number will take you to `/https/domains/#q=Department%20of%20Health%20And%20Human%20Services`, which shows the list of domains filtered to those owned by HHS.

Fixes #101, and #118.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -63,4 +63,5 @@ http://updates.html5rocks.com/2014/11/Support-for-theme-color-in-Chrome-39-for-A
 <script src="/assets/js/vendor/DataTables-1.10.6/media/js/jquery.dataTables.min.js"></script>
 <script src="/assets/js/vendor/DataTables-1.10.6/media/js/dataTables.responsive.js"></script>
 <script src="/assets/js/vendor/d3.min.js"></script>
+<script src="/assets/js/vendor/querystring.js"></script>
 <script src="/assets/js/utils.js?{{ site.time | date: "%Y%m%j%H%M%S" }}"></script>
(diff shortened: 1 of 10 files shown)


juliaelman:
Made one small nitpicky change to from link to filter. I tried out filtering out the agencies listed to the agencies tab, but not entirely sure that made a lot of sense to do. Possibly for another PR, but wanted to get your thoughts on whether or not it's necessary.

konklone:
> I tried out filtering out the agencies listed to the agencies tab, but not entirely sure that made a lot of sense to do. Possibly for another PR, but wanted to get your thoughts on whether or not it's necessary.

Sorry, I don't understand what you mean. Can you explain a bit more?

juliaelman:
With this pull request you are able to go from `analytics/agencies/`, click on the number of domains and get a list of filtered domains by agency at `analytics/domains/#q=`.

My thought was to make the Agency name clickable on `/analytics/domains/` and filter to the Agency tab with that particular agency. See screenshot for where the link would be:

screen shot 2015-05-29 at 11 26 56 am

konklone:
Ooooh. Yes, the thought had occurred to me too. My reaction is that it's nice, but also a lot easier to fit into the DAP table than the crowded HTTPS table, so I hadn't prioritized it.

PR #164, opened 2015-05-29
Full data re-scan and update

konklone:
This performs a full scan and update of the dataset in Pulse, as of May 29th, 2015.

As input, it uses the .gov domain list [as published here](https://github.com/GSA/data/tree/gh-pages/dotgov-domains) on 2015-03-15. This is the same input as our last scan on April 30th, 2015.

It also uses as input a (not yet publicly) updated CSV that the DAP releases, of participating domains/subdomains that have registered at least 1 visit over the prior 2 weeks. In our last April 30th scan, we used a CSV released around March 15th, so the data is fresher by around 2.5 months.

When rounded, the HTTPS participation % remains the same at **31%**. The DAP participation % rises from **33%** to **43%**.

Note that the eligible domains also shifted a bit, as some domains were reachable during the scan that were not before, and some were unreachable that were reachable before.

**Use HTTPS:**

* April 2015 - 370 use HTTPS **/** 1207 total - **30.7%**
* May 2015 - 371 use HTTPS **/** 1192 total - **31.1%**

**Participate in DAP:**

* April 2015 - 265 participate **/** 792 total - **33.5%**
* May 2015 - 342 participate **/** 790 total - **43.3%**

### DIFF
--- a/assets/data/analytics.csv
+++ b/assets/data/analytics.csv
@@ -1,3 +1,3 @@
 status,value
-active,46
-inactive,54
+active,43
+inactive,57
--- a/assets/data/analytics.csv
+++ b/assets/data/analytics.csv
@@ -1,3 +1,3 @@
 status,value
-active,33
-inactive,67
+active,46
+inactive,54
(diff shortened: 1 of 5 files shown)


gbinal:
:+1:

PR #166, opened 2015-05-29
Copy work, HTTPS security disclaimer

konklone:
This makes it more clear that HTTPS isn't about hacking, and links to the [https.cio.gov FAQ](https://https.cio.gov/faq/) for more explanation.

I also added another FAQ answer specifically for domain owners who think we have it wrong. This is in response to an inquiry like that that we've already received and processed.


konklone:
I just added a line about subdomains to the FAQ, as well.

PR #169, opened 2015-05-30
Show more HTTPS, HSTS, and TLS details

konklone:
This expands the extended details we provide for a domain. These only show up on the HTTPS domains table, and require the user to hit the blue `+` button to see the extended information. This information will get merged in with other fields if the table is responsively squashed down on smaller screens.

Example:

screenshot from 2015-05-30 11 45 19

This information is meant to make the table more humane and to provide more detailed explanation for domain owners, and especially to have a place to link to resources that can explain how to address the issues. Right now, only one of the links goes to https.cio.gov (the HSTS one), but this may point to what sections need expanding on https.cio.gov, and how they should be framed (as a fix-it guide, not just an explainer).

The following things about a domain may appear in the HTTP and TLS extended fields:

* Whether it supports or enforces HTTPS.
* Whether it has HSTS enabled, and how strongly.
* Whether its certificate chain has validation issues. This is mostly a lack of intermediate certs, which will work _most_ of the time on _most_ devices, but which should be fixed.
* Whether it supports the deprecated RC4 cipher.
* Whether its (end entity) certificate is signed with SHA-1.
* Whether it supports the deprecated SSLv3 protocol.
* Whether it is missing support for the most modern version of TLS (1.2).
* Whether the domain supports forward secrecy for at least modern clients.

The above fields all appear in the data the table loads, and the sentence/link construction is done in JavaScript.

Some further work I'd like to see:

* A more obvious way to expand the row to see extended details than the blue + button. I'm putting some fairly important stuff here, and it should be obvious to a domain owner that we have more to help them.
* A visual indicator of when someone's doing the right thing. When we say they enforce HTTPS, or are HSTS preload-ready, this should feel different than the admonishments they get for supporting SSLv3 or whatever.


konklone:
Fixed merge conflict (CSS mapping file for some reason), and ready for review.

juliaelman:
Made a few tiny style changes, as well as adding an N/A value to the data if nothing is returned for the TLS issues.

Also, we seem to be running into a lot of merge conflicts, primarily having to do with differing .map files when making changes to the SCSS. My first thought was that having `make watch` running while pulling down newer versions of the code base might be causing the issue. It might be good practice for us to make sure that we turn turn off that process when pulling down new code. Thoughts?

konklone:
> My first thought was that having make watch running while pulling down newer versions of the code base might be causing the issue. It might be good practice for us to make sure that we turn turn off that process when pulling down new code. Thoughts?

Could be. Let's try that and see if it resolves the issue.

PR #170, opened 2015-05-30
Allow redirector domains to "Enforce HTTPS" as long as they redirect to HTTPS somewhere immediately

konklone:
This credits "redirect domains" that always _immediately_ redirect to an HTTPS URL (even if it's another domain's URL) as "Enforcing HTTPS". 

For example, the HTTP endpoints for `18f.gov` and `wh.gov` both always redirect to an HTTPS URL, rather than first redirecting internally. There's no good security reason for the domains to first redirect internally before redirecting externally (it doesn't increase or decrease attack surface for the user), so they should be credited as enforcing HTTPS. They still have the same obligation to add HSTS as everyone else. (And in `18f.gov` and `wh.gov`'s case, they do.)

This affects 8 domains: `nic.gov`, `wh.gov`, `18f.gov`, `bfelob.gov`, `nationalresourcedirectory.gov`, `myra.gov`, `fedbizopps.gov`, and `earmarks.gov`.

This is downstream of #169, which should be merged first.


assets/css/main.css
```diff
@@ -576,7 +576,8 @@ a {
       background-position: center left; }
     .dataTables_wrapper table.dataTable.no-footer {
       border: none; }
-    .dataTables_wrapper table.dataTable.domain td:first-child {
+    .dataTables_wrapper table.dataTable.domain td.odd:first-child,
+    .dataTables_wrapper table.dataTable.domain td.even:first-child {
```
konklone: It looks like the CSS changes here might have been a result of how I addressed the merge conflict earlier in #169. The changes here are correct and should be merged in with this PR.

konklone:
Ah ha, yeah -- there are a bunch of commits from #169 here, because it was downstream. (Duh.) 

The relevant changes are here: https://github.com/18F/pulse/commit/9705ffcba6cd0c7aabd06bd3c2267cf0500406ef

juliaelman:
+1

PR #171, opened 2015-05-30
Mark domains with weak HSTS max-age's as not supporting HSTS

konklone:
This marks a few domains as _not_ using HSTS, when they've set a `max-age` of less than 18 weeks (as the [preload list](https://hstspreload.appspot.com/) requires) and they aren't already known to be preload-ready. (A few domains have weaker max-age's on their canonical `www` endpoint than the strong one on their root domain.)

This affects 3 domains: wh.gov, whitehouse.gov, and ustr.gov.

The OMB policy requires a `max-age` of 1 year, which is more than the preload list. I want to note this, but not punish the (many) domains which are successfully preloaded with max-ages between 18 weeks and 1 year. 

So, I'll be updating this PR to also provide a warning in the extended info for those domains who have max-age's between 18 weeks and 1 year, even though their Yes/No status for HSTS won't change.

The relevant changes so far are: https://github.com/18F/pulse/commit/2b73c8a9e1c91d0af64722304f6d4e405f57c059

This is downstream from #170, which should be merged first.


konklone:
I added the notes on HSTS max ages in the front-end:

screenshot from 2015-05-30 15 47 57

This adds a column to the JSON table file we download, so that the JS can see the max-age and use it to construct the message.

These changes are in https://github.com/18F/pulse/commit/8c71bbe2c884e82107ee2b82c76581168eb9c1b5. This PR is now ready to merge.

konklone:
I made a tweak so that it doesn't say "HSTS is not enabled" followed by a HSTS max-age warning.

screenshot from 2015-05-30 15 51 45

juliaelman:
+1 good addition and adding links to explain more details about the extended information.

PR #173, opened 2015-05-31
Dropdown fix for IE10-11.

### LINKED ISSUE #160: Dropdown menu renders badly in Internet Explorer
In IE9: 
image

...and in Chrome: 
image

juliaelman:
This is a fix for the dropdown to render per the design spec IE 10-11. As for IE9, after doing some digging it looks like this is the best fix >> http://stackoverflow.com/questions/19407332/how-to-hide-drop-down-arrow-in-ie8-ie9 However, this is super hacky and will likely become more trouble than it's worth due to positioning issues of the z-indexed overlay. 

IMHO I feel that it's completely okay to have the dropdown as-is in IE9.

Fixes #160.


konklone:
Awesome, thanks @juliaelman!

PR #174, opened 2015-05-31
Make rows in HTTPS domains table expand when clicked on

konklone:
On the HTTPS domains table, this makes it so that clicking anywhere on the entire row will cause the row to expand to show its details.

It also adds a hover state with a background color, and a mouse pointer cursor, so that it's more obvious that this is possible.

@juliaelman, improvements and alternate suggestions welcome. Screenshot below:

screenshot from 2015-05-31 18 01 24

Downstream of #172, which should be merged first. The relevant work is in https://github.com/18F/pulse/commit/a5365afc5a0041a6241545176168689a92fc4721.

(There's also a tiny language tweak I put in here -- "Yes, and preloaded" -> "Yes, and preload-ready".)


konklone:
(The screenshot tool I used removed the mouse cursor, which was hovering over the highlighted row.)

juliaelman:
Great addition! I did end up changing the color and also making the child row included in that hover state,

konklone:
Beautiful!

PR #175, opened 2015-06-01
Move length control to a list of links

konklone:
This changes the length control from a dropdown to a list of links:

screenshot from 2015-05-31 21 52 46

It uses a `lengthLinks` plugin to DataTables, written by the author. It might need further tweaking to the HTML to get the right attributes in there, but that should be a straightforward modification of the JS, now that we have the foundation in place.

This is downstream of #174, in my unfortunate tradition of making my PRs in serial (once any two of the PRs contains what would otherwise be a terrifying merge conflict).


juliaelman:
Looks good and will follow up with @nickbristow on best approaches to making this accessible.

PR #176, opened 2015-06-01
Always link to the SSL Labs report

konklone:
And remove the "No data." descriptor for cases where we do have data (there just aren't any negative issues to call out, among the ones we look for here).

screenshot from 2015-06-01 01 20 56

I do this in part because the actual grade link is small and easy to miss -- and even harder to click now that we made the whole row clickable, since the cursor pointer is always-on.


juliaelman:
+1

PR #179, opened 2015-06-01
Adding browser warning for anything below IE9.

### LINKED ISSUE #162: consider a browser warning
In the same way that we do on 18f.gsa.gov, I would support including a browser warning that pops up and encourages updating if someone visits with an outdated browser.  It would provide more cover for any negative UI people experience if they visit the site using IE 8 or the like.  

https://github.com/18F/18f.gsa.gov/pull/107

juliaelman:
Fixes #162.

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -6,6 +6,19 @@
 </head>
 
 <body id="{{ page.id }}">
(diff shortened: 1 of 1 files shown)


konklone:
:+1:

PR #180, opened 2015-06-01
Fixing wrapping issue of domain column for longer domain names.

### LINKED ISSUE #161: a few urls overlap the 2nd column
Not urgent but might be worth addressing what happens when a domain is so long it overlaps into the 2nd column.  I'd tend to think have it fall behind the 2nd column and not overlap it.  

screen shot 2015-05-29 at 1 07 21 pm

juliaelman:
Fixes #161


konklone:
Looks good!

PR #181, opened 2015-06-01
Fix/link tabbing

nickbristow:


konklone:
Nice, thank you @nickbristow!

PR #187, opened 2015-06-02
Make pulse.cio.gov the canonical URL

konklone:
This will let search engines prefer our production site, and transfer any PageRank for our staging site to our production site.


juliaelman:
+1

PR #188, opened 2015-06-02
Removed blank links

nickbristow:
Links with no text were appearing in the domain table when their grade wasn't provided for ssl. 

These links are removed


juliaelman:
+1

PR #197, opened 2015-06-02
changes alpha to beta

gbinal:
note that class stays as alpha, since I'm not sure there's a beta tag class yet.


juliaelman:
Looks good! Added a small change of adding a new class name to make more generic.

Issue #203, opened 2015-06-03
HTTPS By Agency data table is not aligned


uncompiled:
The table header does not align with the data rows below them when the window width is greater 1024.

![hi_dpi](https://cloud.githubusercontent.com/assets/408989/7952112/a2010926-097e-11e5-9752-f19d3a840080.png)

In contrast, on a tablet or when the maximum width is less than 1024, the header and the rows are aligned.  :+1: 

![low_dpi](https://cloud.githubusercontent.com/assets/408989/7952123/baf6141c-097e-11e5-991a-1e9525b94980.png)

I noticed the left padding for the rows on the agency tab are removed by the following in main.css:
`.dataTables_wrapper table.dataTable.agency td { padding: 15px 0 !important; }`, so I changed it to: `.dataTables_wrapper table.dataTable.agency td { padding: 15px 0px 15px 40px !important; }` with the following result:

![updated](https://cloud.githubusercontent.com/assets/408989/7952248/e82bea36-0980-11e5-9626-970e628aa5c6.png)

As a side note, the padding exaggerates an issue with the alignment on the progress bars.  If you look at the screenshot, the progress bar for DHHS under the Strict Transport Security column doesn't line up either. It's really only noticeable because 4% is significant enough to be visible, but not wide enough to cover the rounded end of the progress bar.

konklone:
Great catch, thank you! Do you mind preparing a PR with the changes you isolated?

uncompiled:
No problem!

PR #210, opened 2015-06-03
Fix table padding on https/agency tab

uncompiled:
Fixes 18F/pulse#203 by adding left padding in the scss for table
rows on the agency tab.


juliaelman:
+1 thanks for the PR!

PR #212, opened 2015-06-04
Added package.json, cleaned up hookshot.js

ScottONeal:
Thanks for this cool initiative.

Although not used much by contributors, I decided to clean up deploy/hookshot.js, which also involved adding a package.json.

Switched from `minimalist` to `commander` for most robust cli options

hookshot.js now has `--help` command
deploy/hookshot.js --help

Deferred logging of hookshot 'started' line, until the hookshot server was actually listening on the port.

Quite a minor PR, don't mind if it gets reject!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,6 @@ logs
 
 # Mac specific
 .DS_Store
+
+# Node
+node_modules/
--- a/deploy/hookshot.js
+++ b/deploy/hookshot.js
@@ -1,18 +1,20 @@
 #!/usr/bin/env node
 
 var hookshot = require("hookshot");
-var spawn = require("child_process").spawn;
-var options = require('minimist')(process.argv.slice(2));
+var program  = require("commander");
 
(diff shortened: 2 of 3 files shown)


konklone:
Thanks, @ScottONeal! Very happy to accept, though I made a change in https://github.com/18F/pulse/commit/314c7be598c5341cdd319375731ffa580d84b36f to move the `package.json` into the `deploy/` directory, so that it didn't clutter the project root with things that generally aren't used. I also stripped down the `package.json` to the only thing we're using it for, a dependency manifest.

I didn't know about `commander` but it looks pretty nice, and I'll use this for future work. We use this hookshot script for [our main website too](https://github.com/18F/18f.gsa.gov), so we'll migrate it there at some point (unless you beat us to it).

Much appreciated!

ScottONeal:
:beers: I hadn't seen hookshot before, looked pretty neat. I might try it out for a couple projects @uscensus to build an internal CI/CD.

PR #215, opened 2015-06-04
Small typo that jumped out at mme

tswicegood:


juliaelman:
Good catch - thanks @tswicegood!

PR #221, opened 2015-06-12
Data update - June 7, 2015

konklone:
This data update goes from a scan on **May 29, 2015** to a scan on **June 7, 2015**. (There's another scan in progress now, but I'd like to get this merged so we can see granular differences.)

The DAP data has not changed here (and won't be changing in future updates until DAP speeds up its export process). 

The HTTPS number goes up from **31%** to **32%**. A couple notable things:

* sba.gov and sbir.gov are now unreachable by cURL with default options, they get an "Unknown SSL protocol error". SSL Labs doesn't get blocked in the same way, but indicates they both only support TLSv1.0. When I try cURL with `--tlsv1.0`, I can get through to both sites. When I try it with no options, or with `--tlsv1.2`, I get the same unknown error. I think something changed on those domains over that week, for the worst, so they're now marked as not using HTTPS because our scanners can't connect properly. I'm comfortable pushing that back over to them to fix/
* 5 domains went from having good chains to bad chains, and 19 domains went from having bad chains to good chains (including `pay.gov`).

**Use HTTPS:**

* 2015-06-07 - 371 use HTTPS **/** 1192 total - **31.1%**
* 2015-06-07 - 381 use HTTPS **/** 1194 total - **31.9%**

### DIFF
--- a/data/meta.json
+++ b/data/meta.json
@@ -1,0 +1,5 @@
+{
+  "command": "./scan ../data/federal-2015-03-15.csv --scan=inspect,tls,analytics --analytics=../data/dap-2015-05-29.csv --debug --output=../reports/federal-2015-06-07",
+  "end_time": "2015-06-09T18:52:41Z",
+  "start_time": "2015-06-09T18:52:33Z"
+}
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@ logs
 /_site
 /.sass-cache
 /data/*.csv
-/data/*.json
 
 # Mac specific
 .DS_Store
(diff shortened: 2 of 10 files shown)


data/meta.json
```diff
@@ -0,0 +1,5 @@
+{
+  "command": "./scan ../data/federal-2015-03-15.csv --scan=inspect,tls,analytics --analytics=../data/dap-2015-05-29.csv --debug --output=../reports/federal-2015-06-07",
+  "end_time": "2015-06-09T18:52:41Z",
+  "start_time": "2015-06-09T18:52:33Z"
+}
```
konklone: I've added a `meta.json` to memorialize the scan parameters, after adding support in https://github.com/18F/domain-scan/pull/18.

juliaelman:
:+1:

PR #222, opened 2015-06-12
CSV downloads for domain tables

### LINKED ISSUE #72: Data availability and download information
Could be its own page, or its own section, but we should have a place for "Here's the data!" with the JSON and CSV that we have.

konklone:
This adds a CSV download link above the per-domain tables for HTTPS and for DAP:

pasted_image_at_2015_06_12_05_37_pm

The HTTPS CSV includes an `Agency` column, even though the display table does not. The CSV should match the display table exactly, as I have -- in a move that pains my software engineering brain consierably -- copy and pasted the number->value mappings from JavaScript to Python, to ensure they look the same. I think this is pretty important for the audience using these particular CSVs, at least. 

I think it's fair to say I'm unhappy with how many different formats of data are used to build this whole system. In the meantime, here are CSV downloads.

This is downstream of #221, which should be merged first. Fixes #72.


gbinal:
Thanks, @konklone!  This is a great start.

gbinal:
I'm happy to merge.

konklone:
@gbinal You'll need to review and merge #221 first, which is a data update from the site's first week that brings the count from 31 to 32%. Feel free to merge so they go up to staging, but I'd like to resolve some other outstanding issues before we deploy these changes to production.

juliaelman:
Perfect! :+1:

PR #223, opened 2015-06-12
Remove asterisks from sites with bad chains

konklone:
For domains with bad certificate chains, this removes the asterisk, turning `Yes*` to just `Yes`. The asterisk wasn't clearly connected to the note about a bad certificate chain and confused at least one person who wrote in. I'm comfortable leaving the text note + the SSL Labs report link as evidence thereof.

The change is in https://github.com/18F/pulse/commit/29aea883bfb3807e7a6e09627ca23a9a31734af5.

This is downstream from #222, which should be merged first.

PR #224, opened 2015-06-13
Bad bug with logic around bad certificate chains

konklone:
My logic that checked to see if a domain was using a bad certificate chain (which indicates the site is *using* HTTPS, even if it's not fully valid) failed to also make sure the domain's cert wasn't _also_ using a bad hostname (which indicates the site is *not* using HTTPS, whether or not the chain is good).

The fix is in https://github.com/18F/pulse/commit/52f82efd59b94084fe3098ef87c57d9511954190, which brings the number of sites that Use HTTPS down from 381 to 328, out of 1,194 domains. That goes from **31.9%** to **27.4%**, a loss of 4.5%, which makes the big number go from 32 to 27.

A mild silver lining is that a few of those domains actually do use HTTPS, and fixing this bug revealed a smaller bug afflicting those. Fixing that should at least bring this up to something that rounds to 28%, and since the site is currently at 31% (before deploying #221), we can limit the visual damage to 3%. I'm also in the middle of a scan of the last week since #221, which I'm hoping will lead to greater uptake. 

I think the result of all this should be a part of the site, maybe part of the `/about/` page, which has some prose or a list mentioning methodology changes and some progress milestones, for transparency about the changes.

Downstream of #223, which should be merged first.


konklone:
The changes here are in https://github.com/18F/pulse/commit/52f82efd59b94084fe3098ef87c57d9511954190.

juliaelman:
Good catch! Adding a new issue to update `/about/` to outline these changes and inform the user.

PR #225, opened 2015-06-13
Add two fields to download CSVs: branch and redirect status

konklone:
This adds a field for `Branch` and a field for `Redirect` to the download CSVs we provide. We use these two fields to calculate DAP eligibility, and they also will help consumers of the data narrow down the data to what concerns them the most.

The work is in https://github.com/18F/pulse/commit/aebeb5490793f400dd84e7c4398a07173046956c.

Downstream of #224, which should be merged first. (I hate this, but versioning data files basically necessitates this.)


konklone:
The work is in https://github.com/18F/pulse/commit/aebeb5490793f400dd84e7c4398a07173046956c.

juliaelman:
Looks good. Are we eventually wanting to add these as columns for the table? Or just leave in the downloaded CSV for now?

konklone:
They might be helpful as invisible fields for filtering (like Agency is for HTTPS), but I'm not concerned enough to worry about dealing with the extra space. They're just intended to make the CSV better for now.

konklone:
(Merging on the strength of your "Looks good".)

PR #226, opened 2015-06-13
Version the input data from domain-scan after all

konklone:
This versions `inspect.csv`, `tls.csv`, and `analytics.csv`, both for transparency of our inputs, and more clarity during diffs and data updates.

The work is in https://github.com/18F/pulse/commit/34a8f694ea177d94d86a64a8d509593d933191a4, but all it is adding the 3 CSVs we used for the `2015-06-07` update (and `analytics.csv` is the same as for our `2015-05-29` update we launched with).

This is downstream of #225, which should be merged first.


juliaelman:
:+1: new and shiny!

PR #233, opened 2015-06-17
Get DAP off of incorrect central snippet, submit agency/subagency params

konklone:
This fixes the DAP snippet to be in full compliance. It's not using the still-yet-to-be-finalized central hosting location, so I moved it to self-hosted for the time being.


juliaelman:
+1

PR #238, opened 2015-06-19
Fix DAP snippet...again

konklone:
I still messed up the last one in #236, and didn't include the `agency` and `subagency` parameters. Done here.


juliaelman:
+1

PR #244, opened 2015-06-29
Deploy hotfixed scan date to production

konklone:
I neglected to manually update `_config.yml` with the scan date. I hotfixed it on `master` and this deploys it to production.

An idea for future work: during the `./update` process, use the (now-versioned) `meta.json` file to automatically update `_config.yml` with the scan date.


juliaelman:
+1

PR #246, opened 2015-07-03
Data update for July 3, 2015

konklone:
The nicest thing in this update is that **usa.gov now uses and forces HTTPS**, which, as the federal front door, is an awesome thing. There's also a big set of improvements in forward secrecy configuration, which is :+1:.

* Now forcing HTTPS: `usa.gov`, `afrh.gov`, `ecpic.gov`, `nlrb.gov`, `realpropertyprofile.gov`
* Now using HTTPS: `doioig.gov`, `challenge.gov` (`challenge.gov` appears to default to HTTPS, but its redirect path is too tortuous for our scans to pick this up)

A few particularly nice TLS improvements from important sites:

* Good on the FCC: `fcc.gov` turned off RC4 and went from SHA-1 to SHA-2 (B to A)
* Good on the FTC: `ftccomplaintassistant.gov` upgraded SHA-1 to SHA-2, and improved forward secrecy (A- to A)
* Good on NIST: `nist.gov` adds TLS 1.2, better forward secrecy (C to A)

And a raft of TLS improvements generallly:

* Better forward secrecy: `applicationmanager.gov`, `clinicaltrials.gov`, `cpsc.gov`, `dol.gov`, `dot.gov`, `ftc.gov`, `myra.gov`, `osha.gov`, `saferproducts.gov`, `senate.gov`, `sentinel.gov`, `sustainablecommunities.gov`, `transportation.gov`, `treasurydirect.gov`, `usajobs.gov`, `usastaffing.gov`, `ftccomplaintassistant.gov`, `nist.gov`
* Upgraded SHA-1 to SHA-2: `opensource.gov`, `smartgrid.gov`, `uspsoig.gov`, `wrp.gov`, `fcc.gov`, `ftccomplaintasisstant.gov`
* Other improvements: `directoasucuenta.gov` (F to A-), `eta-find.gov` (F to B), fedcenter.gov (C to A-), godirect.gov (F to B)

The overall %'s are not changed.

This update uses fresh HTTPS data from a public scan on 2015-07-3, but still the DAP export from May 2015. However, https://github.com/18F/analytics-reporter/pull/105 has been filed to begin exporting participating second-level DAP domains automatically, which should lead to better updates in that data as well. Hopefully this will be the last scan using static DAP participation data. There are updates to the DAP data, but they only reflect shifts in whether domains are reachable/redirects (and thus eligible for DAP participation at all).


juliaelman:
Looks good! Added the date change to `_config.yml`, but am wondering if we update the domains export information too at this point. Thoughts?

konklone:
Ugh, thanks for remembering the `_config.yml` update. Another great reason to get this stuff fully automated...

> am wondering if we update the domains export information too at this point. Thoughts?

I'm still finalizing this with @tdlowden and would personally rather make that a goal for the next update.

juliaelman:
+1

PR #254, opened 2015-07-25
Simplified data update instructions

konklone:
Now that https://github.com/18F/domain-scan/pull/36 means domain-scan can auto-download the DAP data, the mechanics are substantially simpler.

I'd like to get to the point where the `.gov` domain list is also at a predictable in-place URL. I filed the work to support it in https://github.com/18F/domain-scan/issues/37, but it also requires change at the .gov registry level.


konklone:
This is docs-only, and it's been up a while for review. Merging.

PR #270, opened 2015-08-20
Add configuration for serving production Pulse

mbland:
Requires the production instance be built with:
  bundle exec jekyll serve --config=_config.yml,_config_prod.yml

cc: @juliaelman @konklone

### DIFF
--- a/_config_prod.yml
+++ b/_config_prod.yml
@@ -1,0 +1,1 @@
+baseurl:


konklone:
Is it a problem to put the burden on the staging instance? I'd prefer not to modify production for this, and I'd also like the default development experience to be: run the app, open up `http://localhost:4000`, instead of opening up `http://localhost:4000/pulse/`.

mbland:
It's not possible at the moment; but I'm currently testing and refactoring the Pages server (18F/pages#19), after which it would be possible to add support for, say, a `_config_18f_pages_staging.yml` file, or a `.18f_pages.yml` file with a `staging:` attribute, or something along those lines.

Until that can happen, an alternate workaround that could be done in this repo would be to have a `./go` script like 18F/hub, 18F/dashboard, 18F/18f.gsa.gov, 18F/team-api, etc. Then you could define, for example, `./go serve` and `./go build_prod` or whatever, and redefine them once native 18F Pages support is implemented with no change to the interface for devs or production.

I just pushed a new 18F/go_script repo for the [`go_script` Rubygem](https://rubygems.org/gems/go_script) to make this sort of script easier and more consistent to write and define. I've got a couple small tweaks to make to it which I hope'll go in sometime Monday, but it's perfectly usable right now.

konklone:
Given that 18F Pages always knows it will be hosting a path-prefixed URL, could it just override the value for `baseurl` with the name of the repo automatically? Not sure if it's a CLI flag for `jekyll` anymore, but if not then 18F Pages could maybe just generate its own `.yml` file on the fly to overlay on top of the app's default one.

mbland:
`jekyll server` supports a command-line `--baseurl` parameter. For some reason, `jekyll build` doesn't. It could perhaps generate its own `_config_18f_pages.yml` on the fly. But again, 18F/pages#19 needs to go in first.

mbland:
18F/pages#24 renders this PR moot. Closing.

PR #285, opened 2015-08-30
Data update - August 28, 2015

konklone:
Slow week.

In this update, DAP dipped back down to 44% - 3 domains' participation level changed (`mapaplanet.gov`, `ap.gov`, and `foreignassistance.gov`), and a few other sites became eligible/ineligible by becoming reachable/unreachable, so it's pretty much right on the edge.

**HTTPS/HSTS adoption:**

* Now forcing HTTPS: `isotope.gov` and `isotopes.gov` (the same site), `ornl.gov`

* Now using HTTPS: `fgdc.gov`, `nemi.gov`, `pbgc.gov`

* HSTS: (No changes)

**Improvements to TLS quality:**

* Disabled SSLv3: `fdr.gov`, `federalreserve.gov`
* Upgraded to SHA-2: `dotideahub.gov`
* Removed RC4: `doi.gov`, `eta-find.gov`, `fbiic.gov`, `ffiec.gov`, `symbols.gov`
* Improved forward secrecy: `fbiic.gov``, federalreserve.gov`, `ffiec.gov`, `iad.gov`, `ioss.gov`
* Enabled OCSP stapling: `hru.gov`
* :-1: - `efpts.gov` dropped down to SHA-1. `peacecorps.gov` enabled RC4.


djharrity:
It appears that several active ED websites are showing false: childstats.gov, nationsreportcard.gov,

konklone:
> It appears that several active ED websites are showing false: childstats.gov, nationsreportcard.gov,

@djharrity Er, wow, you're right. We've marked those sites as unreachable for at least the two weeks running.

It's because our scanning server (an EC2 instance in us-east) can't connect to these sites over HTTP:

```
$ curl --head http://www.childstats.gov
curl: (56) Recv failure: Connection reset by peer
$ curl --head http://www.nationsreportcard.gov
curl: (56) Recv failure: Connection reset by peer

```

However, I can access the site via `curl` or by browser on my local machine. I wonder if their servers have blocked our scanning server's IP address. I can also reach them from a random Digital Ocean server.

Thank you for pointing this out. We may need to expand our scanning apparatus to range over multiple IP addresses.

cc @ozzyjohnson @adelevie

konklone:
To be clear, that shouldn't block the merge of this pull request, as this is just the status quo. We'll fix this in a future update, and look for other sites which might be affected in the same way.

konklone:
@djharrity This appears to still be the case about `childstats.gov` and `nationsreportcard.gov`. Do you know anything about what's on Ed's end that might be interfering with our attempts to reach it by `curl` from our public EC2 IP?

djharrity:
@konklone Nope, but will check into it.

PR #291, opened 2015-09-08
Add everykidinapark.gov

konklone:
This manually adds [everykidinapark.gov](https://everykidinapark.gov/), which launched last week. We're still waiting on a formal update from the .gov team, so I'm adding this one by hand until we get a new one.

@khandelwal, would you mind pinging us if/when `everykid.gov` goes live?

### DIFF
--- a/data/domains.csv
+++ b/data/domains.csv
@@ -1,5 +1,6 @@
 Domain Name,Domain Type,Agency,City,State
 18F.GOV,Federal Agency,General Services Administration,Washington,DC
+EVERYKIDINAPARK.GOV,Federal Agency,Department of the Interior,Washington,DC
 ACUS.GOV,Federal Agency,Administrative Conference of the United States,WASHINGTON,DC
 ACHP.GOV,Federal Agency,Advisory Council on Historic Preservation,Washington,DC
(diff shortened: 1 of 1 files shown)


konklone:
(This adds the domain to the scan/eligibility list, but the data for it won't appear until next week.)

khandelwal:
everykidinapark.gov went live September 1st. It's an Interior domain and project, but because of hosting reasons GSA owns the domain (it was officially transferred over). So I don't know how you want to record that - @NoahKunin? 

I think there's a secondary domain too - everykid.gov that was also supposed to be transferred over - but I don't know what happened to that. I'm not planning on turning that on necessarily. I don't really know how to figure out that users don't like typing the whole inapark part!

konklone:
From Pulse's perspective, it's really the program owner who matters, which is DOI. This may be different for the .gov registrar, and so if/when they prepare a formal update of the .gov list, it's possible they list GSA. It'll be interesting to find out, but since Pulse is temporarily filling this in ourselves, I'd prefer DOI as the program owner for the time being.

khandelwal:
Perfect. I just wanted to make sure.

PR #305, opened 2015-09-16
Implement agency and domain landing pages

konklone:
This renders the agency and domain landing pages "for real", so that `/domain/whitehouse.gov` and `/agency/general-services-administration` work and display real data. Data is preprocessed by `data/update` and saved in a couple of new files, `static/data/agencies.json` and `static/data/domains.json`, which are read in by `pulse.py` and held in memory for use in template rendering.

The agency page has its information rendered in text and bar chart form:

screenshot from 2015-09-16 01-43-18

The domain page has had the basic fields implemented, but I left out the "more details" sentence construction for now, as it's a more involved piece of refactoring (which I'm very eager to do). I think (and I think @xtine does as well) that there's more design iteration to do on this page overall.

screenshot from 2015-09-16 01-43-35

We're not trying to call the landing pages final, but they're now integrated into our new structure, with a data pipeline that works for them.


konklone:
For agencies and domains which are not eligible for evaluation under some kind of category (like legislative branch domains/agencies and DAP participation), the relevant blocks are simply hidden. We might want to handle that case more gracefully.

xtine:
LGTM.

PR #323, opened 2015-11-01
Proper deployment for a Python app through fabric

konklone:
This updates our deployment process to use fabric to properly deploy a dynamic gunicorn-based Python app. I've tested it and successfully deployed our staging and production apps with it.

To deploy to staging, go to the `deploy/` directory, have fabric ready in your `$PATH` somehow, and run:

```
fab deploy
```

To deploy to production, run:

```
fab deploy --set environment=production
```

For convenience, I've added `make staging` and `make production` commands from the project's root that will `cd` into `deploy/` and run the right command for you and `cd/` out. So you can actually just do:

```
make staging
```

and

```
make production
```

Since Fabric is (**_still_**) Python 2 only, I typically make a virtualenv locked to Python 2 just to hold `fabric`, called `fab`, and switch into it before deploying, like so:

```
workon fab
make production
```

This PR also documents the above in the README.

### DIFF
--- a/deploy/fabfile.py
+++ b/deploy/fabfile.py
@@ -1,43 +1,123 @@
 import time
-from fabric.api import run, execute, env, cd
+from fabric.api import run, execute, env
 
-"""
-Manage auto-deploy webhooks remotely.
+# Will default to staging, override with "fab [command] --set environment=production"
 
-Production hook:
-
-  forever start -l $HOME/pulse/hookshot.log -a deploy/hookshot.js -p 5000 -b production -c "cd $HOME/pulse/production/current && git pull && bundle exec jekyll build >> $HOME/pulse/hookshot.log"
-  forever restart deploy/hookshot.js -p 5000 -b production -c "cd $HOME/pulse/production/current && git pull && bundle exec jekyll build >> $HOME/pulse/hookshot.log"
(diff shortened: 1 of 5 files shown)


deploy/fabfile.py
```diff
@@ -1,43 +1,123 @@
...
+
+def deploy():
+  execute(checkout)
+  execute(dependencies)
+  execute(make_current)
+  execute(restart)
+  execute(cleanup)
+
+
+
```
konklone: I left the old `fabfile.py` text (which managed the webhook) commented out below. Since converting to a dynamic app, the site doesn't auto-deploy, and the webhooks are inert. We can and should restore them, if we continue to operate the site through EC2, though auto-deploy of a dynamic app via webhook is more annoying than with a static app.

deploy/pulse.conf
```diff
@@ -1,3 +1,12 @@
+# Ports:
+#
+# 3000: production app
+# 5000: production webhook
+# 6000: staging app
+# 7000: staging webhook
+#
+# 4000: https webhook
+
```
konklone: I also updated the Pulse nginx config file to document the ports at use on the box for webhooks, and to catch the on-server change I made a couple weeks ago to fix the redirect behavior for `http://pulse.cio.gov`.

PR #325, opened 2015-11-02
hound

xtine:


konklone:
I just disabled the Hound CI integration, since it posted a bunch of unhelpful comments on my last PR. I'm happy to re-enable if we've been getting value out of it otherwise in the last few months.

PR #329, opened 2015-11-08
Data update - November 6, 2015

konklone:
This data update is complicated, because we got a full .gov second-level domain data update on November 3rd, the first update since March 15th. There was something like 32 new federal .gov domains, and 40 removed federal .gov domains, and then a little bit of changing agency ownership metadata.

I did the first scan without changing the base domain set, to give a good comparison, and got the below as a result. No change to %, and a pretty slow week for HTTPS.

**HTTPS/HSTS adoption:**

* Now forcing HTTPS: `ice.gov`, `restorethegulf.gov`

* Now using HTTPS: 

* HSTS: `18f.gov`, `pic.gov`, `fsapubs.gov` is now HSTS preload-ready

* :-1: - `itap.gov` added HSTS with includeSubDomains -- but with a negative max-age, and is counting down. `edpubs.gov` is almost there, but doesn't have a cert valid for `https://edpubs.gov`.

**Improvements to TLS quality:**

* Disabled SSLv3: `acquisition.gov`, `acus.gov`
* Added TLSv1.2: `moneyfactorystore.gov`
* Removed RC4: `acquisition.gov`, `acus.gov`
* Improved forward secrecy: `acquisition.gov`, `moneyfactorystore.gov`, `protectyourmove.gov`
* :-1: - `faasafety.gov` added RC4


konklone:
In the new domain set, there were ~32 new domains, and 40 removed domains. Most aren't coming out of the gate forcing HTTPS, just:

* `intelligencecareers.gov`, `mgi.gov`, `robodeidentidad.gov`, `vets.gov`

Overall, the shifting numerator and denominator gave us a 1-point % drop for using HTTPS, from 34% to 33%. Oh well! There's also a bunch of new domains to contact about their HTTPS support, who, as new domains, are already supposed to be using HTTPS out of the gate under M-15-13.

konklone:
Thankfully, I was mistaken about the % drop! I had forgotten to update Pulse's local `domains.csv` with the new 2015-11-03 domain copy, and so the `data/update` script was dropping a bunch of bew domain information. I'm fixing this in a pull request now.

djharrity:
ed.gov is now scoring higher.

konklone:
@djharrity Yes it is! That should show up in tomorrow's scan.

PR #331, opened 2015-11-10
Fix missing new domain info on Pulse

konklone:
I had forgotten to update `data/domains.csv` with a local copy of the new .gov domains, and so even though we *scanned* all the domains, we didn't post-process all the domains. This fixes that, and brings the TTPS % back up from **33%** to **34%**.

I hotfixed to `master`, deployed to https://staging.pulse.cio.gov, and this pull request is to `production`.

An example new domain is `anchorit.gov`:

https://staging.pulse.cio.gov/https/domains/#q=anchorit.gov


xtine:
Let me know if there's something I can look for to catch an update error next time.

konklone:
Updating the base domain list is (unfortunately) a pretty rare event, but I don't know of a way to quickly detect it. Hopefully we can automate the base domain update process down the line. :/

PR #339, opened 2015-11-29
Deploy Nov 28th to production

konklone:
Deploys the 2015-11-28 data update (visible at https://staging.pulse.cio.gov) to production.


konklone:
Thanks, @jtexnl! Visible now at https://pulse.cio.gov.

PR #345, opened 2015-12-22
Data update - December 18, 2015

konklone:
The HTTPS % went up again, from **35%** to **36%** -- that's rather terrific progress for a single week, since it went [up from 34% to 35% last week](https://github.com/18F/pulse/pull/343).

Also outstanding: `cfa.gov` is now HSTS preloaded, and `studentloans.gov` has been submitted for HSTS preload review!

**HTTPS/HSTS adoption:**

* Now forcing HTTPS: `doleta.gov`, `fvap.gov`, `medicaid.gov`, `senatecalendar.gov` (a GPO domain), `tsa.gov`, `usich.gov`, `usitc.gov`

* Now using HTTPS: `fueleconomy.gov`

* HSTS: `cfa`.gov is preloaded, `studentloans.gov` added an HSTS header and is in preload review, `medicaid.gov` added a one-year header, `oea.gov` added a one-year header _with subdomains_

**Improvements to TLS quality:**

* Added TLSv1.2: `eftps.gov`
* Upgraded to SHA-2: `digitalgov.gov`, `fueleconomy.gov`
* Removed RC4: `gpo.gov`

PR #346, opened 2015-12-22
Deploy Dec 18th to production

konklone:
Deploys #345 to production.


konklone:
Expediting this one myself, since we're a couple weeks behind and teammates are going on holiday.

PR #348, opened 2015-12-24
Add app/__init__.py to make app discoverable

onceuponatimeforever:
This fixes the bug where "import app" in the python command line returns
a not found error.
Not having __init__.py makes the localdev environment impossible to run
in pulse.py


konklone:
Thanks! I don't think we use `localdev` here, but I don't mind including this, since it's all-upside.

Issue #350, opened 2015-12-24
Long domain names get squished into the next column


ArcTanSusan:
![css_long_name_bug](https://cloud.githubusercontent.com/assets/954858/11992107/bdb600ec-a9d6-11e5-9c1e-676dd1f0f406.png)

konklone:
What browser is this? I can't reproduce on Chrome or Firefox.

onceuponatimeforever:
On Chrome, I repro-ed on page 3 of http://127.0.0.1:5000/https/domains/.

konklone:
I still can't reproduce on Chrome or Firefox, even when I resize the window to be narrow. What device are you using? I should be able to reproduce in Chrome's extensive device emulator in their developer tools.

onceuponatimeforever:
I can repro on a 13'' macbook air laptop on El Capitain version w/ Chrome.

konklone:
So an 18F colleague, @pkarman, kindly attempted to reproduce this on a 13" Macbook Air, with a maximized and non-maximized window, and wasn't able to reproduce it:

<img width="1254" alt="screen shot 2016-01-05 at 10 45 46 am" src="https://cloud.githubusercontent.com/assets/4592/12126689/afb1e670-b3c0-11e5-9067-88bb09a8d05d.png">

<img width="948" alt="screen shot 2016-01-05 at 10 46 18 am" src="https://cloud.githubusercontent.com/assets/4592/12126690/b1011c3a-b3c0-11e5-93d8-c23b5b244011.png">

Is there any more information you can give us? Is it possible you have your zoom options set to non-default values, either for pulse.cio.gov or Chrome-wide (or OS-wide)?

PR #352, opened 2015-12-26
Smaller font-size in total-desc tet

ArcTanSusan:
I couldn't get the font-size css selector to work inside of the ` #content ` selector, so i've pulled it out as its own selector.

Before
------------
pulse_landing_page

After
----------
larger_text_on_homepage


konklone:
I left a comment in #350 asking for more information about the device you're using, so I can reproduce the effect.

PR #354, opened 2015-12-29
Data update - December 26, 2015

konklone:
This is a slow week (not surprisingly). GPO brought some new redirect domains online, and enforce HTTPS for each of them, which is fantastic. The FDIC OIG launched a new HTTPS-only domain as well.

**HTTPS/HSTS adoption:**

* Now forcing HTTPS: `presidentialdocuments.gov`, `congressionaldirectory.gov`, `congressionalrecord.gov`, `fdicoig.gov`, `housecalendar.gov`, `presidentialdocuments.gov`, `uscode.gov`

* Now using HTTPS: `presidentialdocuments.gov`, `congressionaldirectory.gov`, `congressionalrecord.gov`, `fdicoig.gov`, `housecalendar.gov`, `presidentialdocuments.gov`, `uscode.gov`

* HSTS: `fincen.gov` added HSTS and included subdomains, `gsaig.gov` added an HSTS header for its root domain

* :-1: - `pic.gov` lost its HSTS (preload) header.

**Improvements to TLS quality:**

* Added TLSv1.2: `cpnireporting.gov`, `fincen.gov`, `fnal.gov`
* Upgraded to SHA-2: `donotcall.gov`, `globalchange.gov`
* Removed RC4: `gsaig.gov`
* Improved forward secrecy: `citizencorps.gov`, `fnal.gov`, `ftc.gov`, `gsaig.gov`, `llis.gov`, `ustaxcourt.gov`
* Required SNI: `cpnireporting.gov`

PR #349, opened 2015-12-24
Convert tabs to spaces in pulse.py to follow pep8

onceuponatimeforever:
Resolves #253. 

Includes first 2 commits to fix up localdev setup & python formatting (4 spaces, no tabs).

PR #355, opened 2015-12-29
Deploy Dec 26th to production

konklone:
Deploying #354 to production.


konklone:
This also sends #348 and #349 to production, which should have no display impact.

konklone:
Self-merging while the team is largely on holiday.

PR #356, opened 2015-12-31
Adhere to Pylint, add in dev-requirements.txt

ArcTanSusan:


konklone:
Thanks @onceuponatimeforever! However, there are some issues here. 

* I don't wish to rename the `update` file to `update.py`, and would like it renamed back before a merge.
* Running the `update` script from the project root instead of the `data/` subdirectory is fine. However, https://github.com/onceuponatimeforever/pulse/commit/348427e71115a4dbe5e1d9097db893f17e3baaff wasn't only adding indents, it made path changes as well. There's a cleaner view [by adding ?w=1 to the URL](https://github.com/onceuponatimeforever/pulse/commit/348427e71115a4dbe5e1d9097db893f17e3baaff?w=1), but are there any additional changes in https://github.com/onceuponatimeforever/pulse/commit/348427e71115a4dbe5e1d9097db893f17e3baaff?
* Using `dev-requirements.txt` and `pip` to install Fabric unfortunately doesn't make sense here, because Fabric requires Python 2.x and this project requires Python 3.x, even though this project has a Fabric file. Our team uses Fabric to deploy the app, but we're versioning it for our team's convenience -- it's not expected that other teams will use this method, especially given the inconvenience of the version mismatch.
 
To sum up, I'd be happy to take some `pep8`-driven whitespace fixes, and to make `data/update` runnable from the project root. If you can limit the PR to that, I'm happy to merge.

PR #368, opened 2016-01-24
Moving data out of the repo and into an S3 bucket

konklone:
This is step 1 of the multi-step process of automating Pulse' scanning and data publishing processes. It involves some temporary steps that will be undone in a future step. I'm filing this PR as a milestone, and will be branching from this point and filing a separate PR with the next steps later.

This moves the data out of source control (out of `static/data`) and into an S3 bucket (named `pulse.cio.gov`) in 18F's AWS account. In order to avoid CORS issues or using a non-`.gov` domain for the data, I've set up a temporary nginx proxy for `https://pulse.cio.gov/data/` that proxies to `https://s3.amazonaws.com/pulse.cio.gov/live/data/`. 

The previous contents of `https://pulse.cio.gov/static/data/` are now at `https://pulse.cio.gov/data/`, and when this branch is next deployed, the `/static/data/` content will disappear. For now, the directory structure and files and formats are exactly as they were. The front-end has been updated to point to `https://pulse.cio.gov/data/` instead of `/static/data` for data, even in local/staging mode. 

The use of the S3 bucket for live data querying, and the hardcoding of `pulse.cio.gov` into the front-end, will be replaced shortly with a database-driven dynamic response. At that point, the bucket will only be used for bulk/past data, and for reloading the database as necessary. 

Also, I'd like to get the data replicated somewhere else besides this one bucket, as a backup measure. Maybe that's Internet Archive, maybe it's somewhere else. 

In any case, this gets us away from using the git repository as our live/bulk data store, and into a place where we can start using automated methods to update the scan/processed data in-place.

PR #370, opened 2016-01-25
A complete data processing pipeline

konklone:
This extends Pulse's built-in tooling to wrap the existing post-processing script in a larger script that orchestrates:

1. Reading in which `.gov` domain URL to use from `meta.yml` (currently the `2016-01-19` version).
2. Kicking off [`domain-scan`](https://github.com/18f/domain-scan) with the right default parameters, using that `.gov` domain list. The scan data drops into `data/output/scan/results`.
3. Running the post-processing script (previously `data/update`, now `data/processing.py`) to read in the scan data and generate the data the Pulse front-end uses to render data.
4. Uploading the scanned and post-processed data to the production S3 bucket. It's uploaded in-place to the `/live/` directory, as well as to a snapshotted `/archive/` directory with a subdirectory name that uses the date of the scan.

When the `aws` CLI client is configured, and the right environment variables are set so that `domain-scan`, `site-inspector`, and `ssllabs-scan` all fire correctly, causing Pulse to kick off a new scan and update its production data is as simple as running:

```
./data/update
```

The `data/update` script is the system described above, and runs properly out of the root directory of the project.

This is still not _automated_ in the sense that I haven't rigged up any cronjob or automation on the server that would run this script yet. The next step is to update the server to use this pipeline. However, it's not clear yet how we will notify admins if there is an error during the scan.

This work also entailed two (merged) pull requests to `domain-scan`:

* https://github.com/18F/domain-scan/pull/51 - Allow the domain list to be drawn from a URL rather than just on disk. This allows Pulse to just refer to a URL where an appropiate `.gov` domain list can be found, rather than keep a `domains.csv` directly in the repo. `domains.csv` has been removed from the repo, and `meta.yml` now has a link to the domain list, currently in the [`gsa/data`](https://github.com/gsa/data) repo.
* https://github.com/18F/domain-scan/pull/52 - Exit domain-scan with a proper non-0 exit code if there's an error. This is so that Pulse's update pipeline can properly halt midway through if domain-scan throws an error mid-scan.

This PR is still not ready to be deployed, unless we're fine deploying a version of the app where the landing pages are removed. The app won't be at 100% until it uses a local database (probably sqlite or level), and until it can draw from a different data source per-environment (so that dev and staging don't use the production data).

cc @adelevie @juliaelman @xtine


konklone:
Also note, this PR is to the previous PR, #368, not to the `master` branch.

konklone:
This PR is now ready for review.

PR #382, opened 2016-02-05
Total data refactor, and an actual database

konklone:
This, combined with #368 and #370 (which this PR is against), is a feature-complete refactor of the application's data loading and rendering process.

* The app's script which reads data from `domain-scan` CSVs is mostly rewritten. Instead of producing output JSON/CSV files on disk, it now reads it into a [TinyDB](https://tinydb.readthedocs.org/en/latest/) database, which lives on disk in `data/db.json` (not versioned). No other files are written out to disk. 

* TinyDB is a lightweight JSON store. It functions similarly to Mongo in that "it's just JSON", but similarly to SQLite in that "it's just a file". The API is a bit bespoke, but I've captured all direct database access in the `app/models.py` file, behind reasonably named simple functions. So, any future refactor which moves from TinyDB to something else will only need to re-implement those functions, each of which is pretty straightforward.

* The JSON model the DB affords us, combined with Python's and JavaScript's native handling of dicts/objects, means that a lot of code was simplified, and we're no longer repeating/hardcoding display names all over the place. They're generally centered in `app/data.py`, with some repeated once in HTML in the DataTables column header. (This could also probably be changed if desired.)

* The data which used to live in `/data/processed` in #370, and which was versioned in the repo, is now gone. The S3 uploading pipeline will upload the `domain-scan` CSVs and `db.json` to S3, purely as a backup measure and potentially for links to bulk downloads down the line.

* There are now controllers which load data from the database on demand to serve up JSON for DataTables consumption, and CSVs for user download. The CSV construction (which writes out more human-friendly header names and field values) is more computationally expensive than I'd like, and if it becomes a hotly used thing then we could always implement caching via S3 or similar, but I don't think that will be necessary.

* The front-end JS has been refactored to expect dynamically served JSON with the same schema as it is stored in the database. Values are now all native types, so DAP participation values be `true` and `false` instead of `"0"` and `"1"`, and HTTPS values are `0` and `1` (etc) instead of `"0"` and `"1"` (etc).

* Percentage amounts are all calculated dynamically. The database now only stores absolute numbers of total/eligible domains, and the number of domains which match a particular type, and then the server or template will do the division/rounding to produce the %.

* The URLs for downloadable CSVs has changed, for consistency and simplicity. I don't think anyone had wired up any scripts attached to these, especially since they are optimized for humans and not scripts. If they have, and we broke them, then we can address that.

This refactor has not fully paid down the technical debt the rushed mid-2015 schedule created, but it's accomplished a significant % of it, definitely more than half. Thanks for the team's patience as I worked through it -- this is mostly the cost of how quickly we were able to build and launch the site during our first tiny set of sprints in May and June.

This does not fully complete the automation process, but **this PR is ready for merge upon review**, as are #368 and #370. The app should be feature-complete with what's currently running on production.


README.md
```diff
@@ -57,7 +57,7 @@ This will run the fabric command to deploy to production.
 The command to update the data in Pulse and publish it to production is simple:
 
 ```
-./data/update
+python -m data.update
```
konklone: Note the weird syntax for executing the data script. This is done so that the `data/update.py` file can reach "up" and back down into its sibling `app/` directory, which is not otherwise possible in Python 3.

app/models.py
```diff
@@ -0,0 +1,179 @@
...
+this_dir = os.path.dirname(__file__)
+db = TinyDB(os.path.join(this_dir, '../data/db.json'))
+
+# These functions are meant to be the only ones that access the db
+# directly. If we ever decide to migrate from tinydb, that can all be
+# coordinated here.
+
+# Data loads should clear the entire database first.
+def clear_database():
+  db.purge_tables()
```
konklone: The data loading script completely purges and reloads the database from scratch.

app/models.py
```diff
@@ -0,0 +1,179 @@
...
+      for field in CSV_FIELDS['common']:
+        if FIELD_MAPPING.get(field):
+          row.append(FIELD_MAPPING[field][domain[field]])
+        else:
+          row.append(domain[field])
+      for field in CSV_FIELDS[report_type]:
+        row.append(FIELD_MAPPING[report_type][field][domain[report_type][field]])
+      writer.writerow(row)
+
+    return output.getvalue()
```
konklone: The above `to_csv` function is probably the kludgiest part of this refactor, but it generates human-friendly CSVs for domain-level data from the database, without needing to hardcode in HTTPS/DAP-specific logic. All you need to do is add fields to `app/data.py`'s `CSV_FIELDS` dict.

data/domains.csv
```diff
@@ -0,0 +1,1351 @@
+Domain Name,Domain Type,Agency,City,State
```
konklone: `data.domains.csv` is still versioned in the repository. One TODO is to get this out of it, and have the `data/update.py` script download `domains.csv` from a consistent URL before it begins the scanning process.

data/processing.py
```diff
@@ -221,138 +199,134 @@ def load_data():
...
+  return reports
+
+# Go through each report type and add agency totals for each type.
+def update_agency_totals():
+  all_agencies = Agency.all()
+
+  # For each agency, update their report counts for every domain they have.
+  for agency in all_agencies:
+
+    # TODO: Do direct DB queries for answers, rather than iterating.
```
konklone: Right now to calculate/cache totals per-agency, it goes through each agency and then fetches all eligible domains and counts them up in Python code. This should be completely doable with a single database query per-field instead.

data/processing.py
```diff
@@ -221,138 +199,134 @@ def load_data():
...
+  return (domain["live"] == True)
 
-def evaluating_for_https(domain):
+def eligible_for_analytics(domain):
   return (
-    (domain_data[domain].get('inspect') is not None) and
-    (domain_data[domain]['inspect']["Live"] == "True")
+    (domain["live"] == True) and
+    (domain["redirect"] == False) and
+    (domain["branch"] == "executive")
```
konklone: It's expected that any field needed to calculate "eligibility" for any particular report type (including any future ones) will be put on the top-level of the `Domain` object, even if that means it's data that was brought in from `site-inspector` and `inspect.csv`, rather than the domain file itself.

data/update.py
```diff
@@ -47,6 +46,11 @@
 SCANNERS = os.environ.get("SCANNERS", "inspect,tls,analytics")
 ANALYTICS_URL = "https://analytics.usa.gov/data/live/second-level-domains.csv"
 
+# TODO:
+# --date: override date
+# --skip-scan: skip the scanning part (rely on local scan data)
+# --skip-upload: don't bother uploading anything to S3
```
konklone: TODO above: add a command line options parser and support some helpful dev/debugging flags.

deploy/pulse.conf
```diff
@@ -33,24 +33,6 @@ server {
...
-      proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
-      proxy_set_header X-Forwarded-Proto https;
-      proxy_max_temp_file_size 0;
-
-      proxy_connect_timeout 10;
-      proxy_send_timeout    30;
-      proxy_read_timeout    30;
-
-  }
-
```
konklone: This proxy to the Pulse S3 bucket is no longer necessary.

PR #387, opened 2016-03-07
Better automation tools

konklone:
This does a few things:

* Versions a `deploy/crontab` file in the repository, and creates a fab task to set its contents as the crontab for the deployed-to user.
* Adds `make` commands to wrap the data update process in staging, production, and development (for convenience).
* Adds a `deploy/cron.sh` script that, given a `PULSE_ENV` of either `staging` or `production`, will run either the staging or production data update processes.
* Uses the scan date from `meta.json` (when `domain-scan` was run) rather than the date the data update script is being run. While these will usually be the same in production, it's more correct to use `meta.json` then, and will guarantee the date is correct in other environments where the two dates are likely to be different.
* Gets rid of the `.gov` export date info in `meta.yml`.
* Updates the donuts to not show `NaN` during a data update.
* Adds an RSS feed to `/data/reports/feed/` so that it's possible to automatically follow data update events (or potentially their unplanned absence).


Makefile
```diff
@@ -4,13 +4,13 @@ css ?= static/css/main.css
...
-	cd deploy/ && fab deploy && cd ..
+	cd deploy/ && fab deploy --set environment=staging && cd ..
 
 production:
 	cd deploy/ && fab deploy --set environment=production && cd ..
 
-run:
-	python pulse.py
+debug:
+	DEBUG=true python pulse.py
```
konklone: If the app is run with `DEBUG=true`, it will use Flask's debug mode (and show stack traces, etc.).

app/helpers.py
```diff
@@ -1,22 +1,34 @@
...
 
+  def scan_date():
+    return models.Report.report_time(models.Report.latest()['report_date'])
+
   # Make site metadata available everywhere.
   meta = yaml.safe_load(open("meta.yml"))
   @app.context_processor
   def inject_meta():
-      return dict(site=meta, now=datetime.datetime.utcnow)
+      return dict(site=meta, now=datetime.datetime.utcnow, scan_date=scan_date())
```
konklone: This means that every web request triggers at least one database query, to get the latest scan date.

app/helpers.py
```diff
@@ -1,22 +1,34 @@
...
 
   @app.template_filter('date')
   def datetimeformat(value, format='%H:%M / %d-%m-%Y'):
       return value.strftime(format)
 
-  # TODO: consolidate data mappings in one place.
-  # @app.template_filter('field_map')
-  # def field_map(value, category=None, field=None):
-  #     return data.data[category][field][str(value)]
```
konklone: This function was unused.

data/domains.csv
```diff
@@ -26,6 +26,7 @@ ABILITYONE.GOV,Federal Agency,Comm for People Who Are Blind/Severly Disabled,Arl
 JWOD.GOV,Federal Agency,Comm for People Who Are Blind/Severly Disabled,Arlington,VA
 CFTC.GOV,Federal Agency,Commodity Futures Trading Commission,Washington,DC
 SMARTCHECK.GOV,Federal Agency,Commodity Futures Trading Commission,Washington,DC
+WHISTLEBLOWER.GOV,Federal Agency,Commodity Futures Trading Commission,Washington,DC
```
konklone: This `domains.csv` update to keep up with the `2016-01-19` data update, but is temporary -- we should be able to move to pulling this data right from GitHub, and/or Data.gov, before long.

data/update.py
```diff
@@ -43,38 +45,65 @@
...
-SCANNERS = os.environ.get("SCANNERS", "inspect,tls,analytics")
-ANALYTICS_URL = "https://analytics.usa.gov/data/live/second-level-domains.csv"
-
-# TODO:
-# --date: override date
-# --skip-scan: skip the scanning part (rely on local scan data)
-# --skip-upload: don't bother uploading anything to S3
-
-def run():
+SCANNERS = os.environ.get("SCANNERS", "inspect,tls,analytics,sslyze")
```
konklone: This re-adds `sslyze` to the list of scanners run in production. We don't surface this data in Pulse directly yet, but it's very good to be collecting it along the way..

data/update.py
```diff
@@ -43,38 +45,65 @@
...
+ANALYTICS_URL = os.environ.get("ANALYTICS_URL", META["data"]["analytics_url"])
+
+# Options:
+# --date: override date, defaults to contents of meta.json
+# --scan=[skip,download,here]
+#     skip: skip all scanning, assume CSVs are locally cached
+#     download: download scan data from S3
+#     here: run the default full scan
+# --upload: upload scan data and resulting db.json anything to S3
```
konklone: This adds a very tiny CLI options parser, to support a few options that help distinguish between production, staging, and development use cases for data updating.

deploy/fabfile.py
```diff
@@ -84,50 +91,15 @@ def deploy():
...
-#   run(
-#     "cd %s && forever start -l %s -a deploy/hookshot.js -p %i -b %s -c \"%s\""
-#     % (current, log, port, branch, command)
-#   )
-
-# def stop():
-#   run(
-#     "cd %s && forever stop deploy/hookshot.js -p %i -b %s -c \"%s\""
-#     % (current, port, branch, command)
-#   )
```
konklone: This used to be the fabfile code to manage the webhook, back in the pre-Flask days.

deploy/pulse.conf
```diff
@@ -71,10 +80,11 @@ upstream staging_pulse {
 }
 
 server {
+  listen 443 ssl http2;
   server_name  staging.pulse.cio.gov;
 
-  ssl_certificate      /etc/nginx/ssl/keys/staging.pulse.cio.gov.chained.crt;
-  ssl_certificate_key  /etc/nginx/ssl/keys/staging.pulse.cio.gov.key;
+  ssl_certificate      /etc/letsencrypt/live/pulse.cio.gov/fullchain.pem;
+  ssl_certificate_key  /etc/letsencrypt/live/pulse.cio.gov/privkey.pem;
```
konklone: This re-syncs the Pulse nginx config with what it is on the server, after re-updating the certificate (after the hard drive fault we suffered a month back) to use one from Let's Encrypt.

meta.yml
```diff
@@ -9,11 +9,8 @@ description: "How federal government domains are meeting best practices on the w
...
-  # The scan date.
-  scan: 2016-02-12
-
   # URL where we download the .gov domain list from.
   domains_url: https://raw.githubusercontent.com/GSA/data/gh-pages/dotgov-domains/2016-01-19-federal.csv
 
-  # The export date for the .gov domain list. Increasingly non-relevant.
-  domains: 2015-11-03
+  # URL where we download the participating DAP .gov list from.
+  analytics_url: https://analytics.usa.gov/data/live/second-level-domains.csv
```
konklone: Moves the URL to the DAP data from `data/update` to `meta.yml`.

static/js/analytics/agencies.js
```diff
@@ -36,11 +36,12 @@ $(document).ready(function () {
...
-            if (type == "sort")
-              return null;
-            return Utils.progressBar(Utils.percent(
+            var percent = Utils.percent(
               row.analytics.participating, row.analytics.eligible
-            ));
+            )
+            if (type == "sort")
+              return percent;
+            return Utils.progressBar(percent);
```
konklone: This fixes a bug reported to us by a user -- it was a regression from the refactor, where the rendered percent columns on the Agency pages for each category were not sorting correctly.

konklone:
OK, without objection...

PR #388, opened 2016-03-13
Remove domains.csv from version control, use the one used in scan

konklone:
This removes `data/domains.csv` from version control. Instead, the data loading process will rely on the version of `domains.csv` that was used during the scan. This will generally have been downloaded to  `data/output/scan/cache/domains.csv` as part of the `domain-scan` process. 

However, if this file/directory does not exist (for example, if you're running a data loading process from scan result data downloaded from some prior run, without having actually just run a scan locally), then the data loading script will create the directory and `wget` the domains CSV into the right place, using the same URL in `meta.yml` that a scan would have been instructed to use.


konklone:
This is the last bit of versioned data we were leaning on. Unfortunately, we are still using permalinks that have dates in them. We've been doing this in the gsa/data repo, in part because that is still meant to be an interim location for the data. 

However, the official URL for the data is also fragile. As of right now:

https://inventory.data.gov/dataset/b2c31002-8a5e-4cd0-85bd-6971934f4e59/resource/02706c7b-98e3-4267-8dd8-e3bb2d8c7ce3/download/fulldomainreport01192016.csv

So right now, when there's a data update, we again have to update the URL for this file by hand in `meta.yml`. While that's better than updating the data file, that's still a brittle process and not a good API. I'm considering consolidating the files in `gsa/data` to also create a date-less permalink that points to the "latest" domains, but I would prefer data.gov provide a steady URL instead. cc @philipashlock if he has any thoughts.

konklone:
Without objection...

PR #393, opened 2016-03-18
Update accessibility page

nickbristow:
Updating the accessibility page as discussed


gbinal:
Looks good to me.  Let's add.

.gitignore
```diff
@@ -8,6 +8,7 @@ logs
 
 # virtualenv
 .Python
+.python-version
```
konklone: Were you having issues with this, @nickbristow?

konklone:
:+1:, but a note here that we should remember to remove the `static/data/` file this PR added later. That directory was removed in prior PRs during the refactor.

.gitignore
nickbristow: @konklone I installed pyenv and it added that file, I assumed we didn't need that on github so I added to .gitignore

PR #434, opened 2016-05-19
Incorporating updates

jtexnl:
Deploying


templates/https/agencies.html
```diff
@@ -1,33 +1,37 @@
 {% extends "layout.html" %}
 
 {% block page_id %}https{% endblock %}
-{% block title %}HTTPS in the federal government - By agency{% endblock %}
+{% block title %}Secure HTTP (HTTPS) - Pulse{% endblock %}
```
konklone: Sorry to catch this late, but why did the <title> get renamed here? `HTTPS` is much preferred to `Secure HTTP (HTTPS)`. No one uses the phrase "Secure HTTP".

konklone:
I don't like the HTTPS title change in 20a06b61c6701464b88d88e69880fbd2c83aef80, but I'm not going to block the merge here -- but I would like to discuss it and potentially revert it in a future PR.

PR #440, opened 2016-05-31
Anchor tag scroll on a11y page

### LINKED ISSUE #416: ensure that anchor tags work on individual error pages
When I click on an error category on the domains page, it's always taking me to the top of the individual error page.  I could be wrong but it seems like there aren't `ID=` or similar for the anchor tags to work.  

For instance:  https://pulse.cio.gov/accessibility/domain/aapi.gov#ColorContrastErrors

xtine:
A11y pages with a url hash will load with the error div scrolled to the top:

screen shot 2016-05-31 at 11 26 27 am

Fixes #416

### DIFF
--- a/static/js/accessibility/a11y.js
+++ b/static/js/accessibility/a11y.js
@@ -41,6 +41,13 @@ $(document).ready(function () {
       $element.html('No results found for ' + domain);
     }
 
+  }).done(function () {
+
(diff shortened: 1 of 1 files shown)


gbinal:
Looks good to me.

konklone:
:+1:

PR #452, opened 2016-07-04
Small improvements to production scanning and data archiving

konklone:
This updates the production scanning process to do a few new things:

* Saves the very detailed `cache/` data, which includes per-domain extended JSON data from each scan, alongside the other scan data.
* Tells Amazon S3 to sync from `live/` to `archive/` within S3 itself, rather than re-copying everything from disk to S3 to do so. This became more important once `cache/` was being synced, since that is a more expensive operation.
* Puts the `tls` (SSL Labs) scanner at the end, since it takes the longest. It's easier to see if anything goes wrong with the others if they happen first.

### DIFF
--- a/app/views.py
+++ b/app/views.py
@@ -88,10 +88,10 @@ def register(app):
     # def a11ydomain(hostname=None):
     #   return render_template("a11y.html", domain=hostname)
 
-    # # Sanity-check RSS feed, shows the latest report.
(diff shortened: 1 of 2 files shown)


gbinal:
@konklone - I'm game to merge and include this in the next push.  That okay with you?

konklone:
Yes, I've tested this code in the production environment and it works as expected.

PR #462, opened 2016-07-26
508 Checklist Fixes

xtine:
A few accessibility errors fixed:
- fixed a typo on `tabindex=1` for skiplinks so focus won't be lost
- made sure all main layout links are white to maintain color contrast ratio (specifically for footer)


nickbristow:
👍

PR #466, opened 2016-07-27
Updated datatables.js to the latest version

nickbristow:
Updated datatables responsive to the latest version
Updated first row in each table to use TH instead of TD to be html5
Changed datatables scss to target TH instead of td


xtine:
This is awesome!! 👍

PR #480, opened 2016-08-03
Fixes several 508 issues

nickbristow:
Adds hover state to links that also applies to keyboard focus
Adds aria labels to the expanding rows
  Labels change when expanded and closed
  auto focuses on child rows
  adds aria-expanded attribute to rows
Adds aria-labels to the pagination links
Updated captions
  Screen readers should read table captions now
  Added additional information to table captions


gbinal:
Great work - thanks, @nickbristow!

PR #485, opened 2016-08-08
add in non error domains

nickbristow:


gbinal:
Looks good.  Merging.

Issue #495, opened 2016-08-11
normalize the guidance tabs 


gbinal:
The guidance tabs for both [HTTPS](https://pulse.cio.gov/https/guidance/) and [DAP](https://pulse.cio.gov/analytics/guidance/) both have a clean, consistent, helpful blurb at the top.  

![screen shot 2016-08-11 at 3 39 59 pm](https://cloud.githubusercontent.com/assets/633088/17602265/e26da124-5fd9-11e6-951b-9b41649266be.png)

I'd suggest normalizing the guidance tab for the accessibility section to match.

gbinal:
Addressed by #496

PR #496, opened 2016-08-11
suggested edits to the accessibility guidance tab

gbinal:


gbinal:
Addresses #495

gbinal:
I'm going to go and self-merge this since it's so tied in with making more clear the 'what is measured' question.

PR #513, opened 2016-08-21
Swap out site-inspector for pshtt

konklone:
This uses @dhs-ncats' [`pshtt`](https://github.com/dhs-ncats/pshtt) tool for HTTP scanning. It's meant to be an (almost) drop-in replacement for `site-inspector`. However, it is a standalone tool that only does this one function, it is written in Python (2.X, [for now](https://github.com/dhs-ncats/pshtt/issues/4)), and sees use by others in the federal government.

There are no visible changes to the data, and this should result in nearly the same determination, though the README is different and I swapped out a link in the HTTPS guide page to link to `pshtt` instead.

While right now `site-inspector` and `pshtt` give nearly identical results, there are some differences. One example is that `pshtt` does not grant "HSTS Preload" status to domains which don't have the `include_subdomains` flag turned on in the Chrome HSTS preload list (e.g. `healthcare.gov`). This difference is currently obscured right now on display because Pulse won't display that a domain is preloaded unless it's preload-ready, but this will change in a near-term PR (since Chrome is now auto-removing domains that drop the header, like Firefox).

cc @garrettr @gbinal


konklone:
Ugh, wrong branch. Filing again.

PR #514, opened 2016-08-21
Swap out site-inspector for pshtt

konklone:
This is to the `production` branch, because the `staging` branch is in a state of flux while accessibility work is done. (I may soon do a "catch up" merge of `staging` into `production` and then re-hide the accessibility URLs, so that the branches don't get far afield.)

This uses @dhs-ncats' [`pshtt`](https://github.com/dhs-ncats/pshtt) tool for HTTP scanning. It's meant to be an (almost) drop-in replacement for `site-inspector`. However, it is a standalone tool that only does this one function, it is written in Python (2.X, [for now](https://github.com/dhs-ncats/pshtt/issues/4)), and sees use by others in the federal government.

There are no visible changes to the data, and this should result in nearly the same determination, though the README is different and I swapped out a link in the HTTPS guide page to link to `pshtt` instead.

While right now `site-inspector` and `pshtt` give nearly identical results, there are some differences. One example is that `pshtt` does not grant "HSTS Preload" status to domains which don't have the `include_subdomains` flag turned on in the Chrome HSTS preload list (e.g. `healthcare.gov`). This difference is currently obscured right now on display because Pulse won't display that a domain is preloaded unless it's preload-ready, but this will change in a near-term PR (since Chrome is now auto-removing domains that drop the header, like Firefox).

cc @garrettr @gbinal @h-m-f-t

### DIFF
--- a/data/processing.py
+++ b/data/processing.py
@@ -4,7 +4,7 @@
 #
 # * domains.csv - federal domains, subset of .gov domain list.
 #
-# * inspect.csv - domain-scan, based on site-inspector
+# * pshtt.csv - domain-scan, based on pshtt
 # * tls.csv - domain-scan, based on ssllabs-scan
 # * analytics.csv - domain-scan, based on analytics.usa.gov data
 #
@@ -62,7 +62,7 @@ def run(date):
   for domain in analytics_ineligible:
     analytics_ineligible_map[domain] = True
 
-  # Pull out a few inspect.csv fields as general domain metadata.
+  # Pull out a few pshtt.csv fields as general domain metadata.
   for domain_name in scan_data.keys():
     analytics = scan_data[domain_name].get('analytics', None)
     if analytics:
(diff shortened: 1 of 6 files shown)


konklone:
cc @h-m-f-t

data/processing.py
```diff
@@ -463,11 +462,11 @@ def https_report_for(domain_name, domain, scan_data):
 
     else:
       # This kind of "Partial" means `includeSubdomains`, but no `preload`.
-      if (inspect["HSTS All Subdomains"] == "True"):
+      if (pshtt["HSTS Entire Domain"] == "True"):
```
konklone: This field name changed from `HSTS All Subdomains` to `HSTS Entire Domain`.

data/processing.py
```diff
@@ -62,29 +62,28 @@ def run(date):
...
     else:
-      # print("[%s] Updating with inspection metadata." % domain_name)
-      domains[domain_name]['live'] = boolean_for(inspect['Live'])
-      domains[domain_name]['redirect'] = boolean_for(inspect['Redirect'])
-      domains[domain_name]['canonical'] = inspect['Canonical']
-
+      # print("[%s] Updating with pshtt metadata." % domain_name)
+      domains[domain_name]['live'] = boolean_for(pshtt['Live'])
+      domains[domain_name]['redirect'] = boolean_for(pshtt['Redirect'])
+      domains[domain_name]['canonical'] = pshtt['Canonical URL']
```
konklone: This field name changed from `Canonical` to `Canonical URL`.

data/update.py
```diff
@@ -46,7 +46,7 @@
 # domain-scan information
 SCAN_TARGET = os.path.join(this_dir, "./output/scan")
 SCAN_COMMAND = os.environ.get("DOMAIN_SCAN_PATH", None)
-SCANNERS = os.environ.get("SCANNERS", "inspect,analytics,sslyze,tls")
+SCANNERS = os.environ.get("SCANNERS", "pshtt,analytics,sslyze,inspect,tls")
```
konklone: I'm leaving the `inspect` scanner in the production environment and scan process, so that we still have `inspect.csv` and `cache/inspect/` files saved and uploaded to S3 for a while, to compare as necessary. If there are no problems after a month or two, we can drop `inspect` from the scan workflow. It shouldn't add much time to the overall scan process.

PR #515, opened 2016-08-22
Improve branch detection logic to keep up with .gov list

konklone:
This fixes the detection of legislative and judical branch domains. This will remove legislative and judicial branch domains more effectively for scans where they are not eligible (such as DAP compliance).

### DIFF
--- a/data/processing.py
+++ b/data/processing.py
@@ -640,11 +640,18 @@ def branch_for(agency):
     "Library of Congress",
     "The Legislative Branch (Congress)",
     "Government Printing Office",
-    "Congressional Office of Compliance"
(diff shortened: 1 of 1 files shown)


data/processing.py
```diff
@@ -640,11 +640,18 @@ def branch_for(agency):
     "Library of Congress",
     "The Legislative Branch (Congress)",
     "Government Printing Office",
-    "Congressional Office of Compliance"
+    "Government Publishing Office",
```
konklone: The .gov list doesn't yet say `Government Publishing Office`, but I think they will on their next update. Better to future-proof this.

PR #517, opened 2016-08-22
Breaks out preloaded and HSTS columns

konklone:
This breaks out a domain's HSTS status and preloaded status. It looks like this:

screenshot from 2016-08-21 21-31-40

And this:

screenshot from 2016-08-21 21-32-23

This is for a few reasons:

* To call more attention to preloading, and a domain's lack of it (even though it's not required by M-15-13 [though it is [recommended](https://https.cio.gov/guide/#options-for-hsts-compliance)]).
* To call more attention to domains which are preload-ready, but not yet submitted. (This now says `Ready for submission` if it's ready to go.)
* HSTS is calculated on a per-endpoint basis, where preloading is necessarily a domain-wide calculation. They are measuring two different scopes, and mixing them is confusing.
* To allow for domains to be preloaded without actually being "preload ready". This would happen if the Chrome list makes a manual override, which could happen for several reasons. The Chrome list has about 25 of these right now, including `login.gov`.

This adds another column to the table, but it collapses on a narrow window just like the HSTS field does, so this shouldn't cause any issues.

### DIFF
--- a/app/data.py
+++ b/app/data.py
@@ -14,6 +14,7 @@ LABELS = {
     'uses': 'Uses HTTPS',
     'enforces': 'Enforces HTTPS',
     'hsts': 'Strict Transport Security (HSTS)',
+    'preloaded': 'Preloaded (recommended)',
     'grade': 'SSL Labs Grade',
 
     'hsts_age': 'HSTS max-age',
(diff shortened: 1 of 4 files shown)


konklone:
A few more notes:

* The preloading column is not shown or aggregated in the `Agencies` tab, since agencies are not required to preload -- and because it is not a reasonable thing to expect agencies to hit 100% on by the end of 2016.
* The downloadable CSV is updated with the `Preloaded (recommended)` column, with the same display as in the table.
* The overall .gov-wide HSTS % (not shown, but calculated) comes out to the same number.

PR #518, opened 2016-08-22
Overhaul HTTPS table to always show the next step

konklone:
This overhauls the HTTPS table to:

* Remove the show/hide button that showed extra details about a domain.
* Always show some message about each domain.
* Each message focuses on a domain owner's next step.
  * Though if they have an F grade from SSL Labs, we'll also point that out.

While this makes the table take up more vertical space, I think it's very worthwhile, especially with the rewritten message structure. Removing the show/hide button also let me cut out a bunch of code around making that work and around making it accessible.

I've also updated DataTables (and its Responsive plugin) to their latest versions, and cleaned up the include paths a bit.

The principles behind the message choices are:

* Only grant "perfect score!" if TLS quality issues are gone.
* Don't show TLS quality issues when pushing to preload.
* All flagged TLS quality issues should be reflected in the SSL Labs grade, so that agencies have fair warning of issues even before we show them.
* Don't speak explicitly about M-15-13, since not all domains are subject to OMB requirements.

A screenshot:

screenshot from 2016-08-22 02-08-33

Another screenshot:

screenshot from 2016-08-22 02-14-26

And a third screenshot:

screenshot from 2016-08-22 02-16-46

cc @h-m-f-t @gbinal


konklone:
Also, this is downstream from #517, and contains everything in that PR. So, #517 should be merged (or at least reviewed) first.

static/js/https/domains.js
```diff
@@ -67,7 +71,7 @@ $(document).ready(function () {
       return ""
     else
       return "" +
-        "<a href=\"" + labsUrlFor(row.domain) + "\" target=\"blank\">" +
+        "<a href=\"" + labsUrlFor(row.canonical) + "\" target=\"blank\">" +
```
konklone: SSL Labs has changed their behavior to no longer auto-detect whether a domain is meant to be accessed at `www` or the root of the domain. Since we go to great lengths to figure that out during our `inspect` / `pshtt` scans, we can use that data here to send SSL Labs the endpoint we wish it to scan.

static/js/https/domains.js
```diff
@@ -79,118 +83,168 @@ $(document).ready(function () {
...
+
+      if (row.https.sig == "SHA1withRSA")
+        tls.push("Certificate uses a " + l("sha1", "weak SHA-1 signature"));
+
+      if (row.https.ssl3 == true)
+        tls.push("Supports the " + l("ssl3", "insecure SSLv3 protocol"));
+
+      if (row.https.tls12 == false)
+        tls.push("Lacks support for the " + l("tls12", "most recent version of TLS"));
+    }
```
konklone: I removed the information messages for `RC4` and for forward secrecy. 

RC4 is going to be shut off by browsers regardless, and the only sites that have to worry are sites which **only** use RC4, which I don't think exists in (federal) `.gov`.

Forward secrecy is important, but now that we're gating "Perfect score!" on TLS messages, I no longer think it's worth calling out specifically over it. Since we still gate on A+, SSL Labs can choose to enforce that if it wants (I don't think it holds up an A+ over it, according to their [Rating Guide](https://www.ssllabs.com/downloads/SSL_Server_Rating_Guide.pdf)).

static/js/https/domains.js
```diff
@@ -79,118 +83,168 @@ $(document).ready(function () {
...
+
+    // Though not found through SSL Labs, this is a TLS issue.
+    if (https == 1)
+      tls.push("Certificate chain not valid for all public clients. See " + l(labsUrlFor(row.canonical), "SSL Labs") + " for details.");
```
konklone: Right now this just flags all certificate chain errors. Very soon, I'd like to start surfacing expired certificate warnings in a different way, since we have that data now in `pshtt`. It would also be ideal to distinguish incomplete chains from untrusted roots, but I'm not sure offhand how we'd do that. `pshtt` does not currently make that distinction.

static/js/https/domains.js
```diff
@@ -79,118 +83,168 @@ $(document).ready(function () {
...
     hsts: "https://https.cio.gov/hsts/",
     sha1: "https://https.cio.gov/technical-guidelines/#signature-algorithms",
     ssl3: "https://https.cio.gov/technical-guidelines/#ssl-and-tls",
-    tls: "https://https.cio.gov/technical-guidelines/#ssl-and-tls",
-    fs: "https://https.cio.gov/technical-guidelines/#forward-secrecy",
-    preload: "https://https.cio.gov/hsts/#hsts-preloading"
+    tls12: "https://https.cio.gov/technical-guidelines/#ssl-and-tls",
+    preload: "https://https.cio.gov/hsts/#hsts-preloading",
+    stay_preloaded: "https://hstspreload.appspot.com/#continued-requirements",
+    submit: "https://hstspreload.appspot.com"
```
konklone: For preload-ready domains, this now links directly to the submission page.

static/js/https/domains.js
```diff
@@ -79,118 +83,168 @@ $(document).ready(function () {
...
 
-  var detailsKeyboardCtrl = function(){
-      $('table tbody tr td:first-child').attr('tabindex','0')
-      .attr('aria-label','Select for additional details')
-      .on('keydown',function(e){
-        if (e.keyCode == 13)
-          $(this).click();
-          $(this).parent().next('tr.child').focus();
-      })
-    };
```
konklone: Since the show/hide control is gone, this keyboard control is no longer necessary.

static/js/https/domains.js
```diff
@@ -79,118 +83,168 @@ $(document).ready(function () {
...
     var table = $("table").DataTable({
 
-      responsive: true,
-
       data: data,
 
+      responsive: {
+          details: {
+              type: "",
+              display: $.fn.dataTable.Responsive.display.childRowImmediate
```
konklone: This option is what auto-shows the child row in the Responsive plugin.

static/js/https/domains.js
```diff
@@ -214,16 +268,16 @@ $(document).ready(function () {
...
+          data: "https.preloaded",
+          render: display(names.preloaded)
+        },
+        {
           data: "https.grade",
           render: linkGrade
         },
         {
-          data: "Details",
+          data: "",
```
konklone: No need to show `Details` in front of every message. I'd prefer to get rid of the indentation too, but am not sure how to do that right now.

static/js/https/domains.js
```diff
@@ -250,31 +304,6 @@ $(document).ready(function () {
...
-    });
-
-
-    //Adds keyboard control to first cell of table
-    detailsKeyboardCtrl();
-
-    table.on("draw.dt",function(){
-       detailsKeyboardCtrl();
-    });
-
```
konklone: The click event spreader and the keyboard control functions are no longer necessary now that the detail row auto-shows.

templates/includes/head.html
```diff
@@ -34,18 +34,26 @@
...
+<!-- sigh, datatables -->
+
+<link rel="stylesheet" type="text/css" href="/static/DataTables-1.10.12/css/jquery.dataTables.min.css"/>
+<link rel="stylesheet" type="text/css" href="/static/Responsive-2.1.0/css/responsive.dataTables.min.css"/>
+
+<script type="text/javascript" src="/static/DataTables-1.10.12/js/jquery.dataTables.min.js"></script>
+<script type="text/javascript" src="/static/Responsive-2.1.0/js/dataTables.responsive.min.js"></script>
+
+<script src="/static/js/dataTables.lengthLinks.js"></script>
+<script src="/static/js/dataTables.downloads.js"></script>
```
konklone: The `lengthLink` and `downloads` plugins for DataTables continue to work perfectly even after DataTables is upgraded.

anand-bhat:
When there are trust issues and the site has HSTS enabled, I recall Chrome showing an error describing  the reason for the trust issue but with no option to bypass the error/ visit the site. This may prove to be a roadblock when accessing sites that score a T (e.g., acquisition.gov ) or sites that score an F but have a trust issue (e.g., ipcc-wg2.gov) if they enable HSTS before addressing the trust issue.

Showing "Almost: Enable HSTS so that clients can enforce HTTPS." as the next step may make the site unoperational in browsers that enforce this constraint.

konklone:
Great point. We do actually have a flag for whether the site has an untrusted certificate chain, and SSL Labs has a T grade for sites with an untrusted certificate. We should encourage that those be fixed before pushing to HSTS.

konklone:
cc @djharrity

konklone:
@anand-bhat I added a case for this situation:

screenshot from 2016-08-22 14-52-00

Eventually, we'll want to only show this for sites which serve an incomplete or malformed chain, and not for sites which chain to a non-publicly-trusted root. Agencies can turn on HSTS for a non-publicly-trusted root (such as the FPKI or DoD), and they should if their non-public user base all trusts that root and that's what they want to stick with. However, sweeping that up in this case is fine for now.

Bad hostname errors are not addressed here, because a site with a certificate for the wrong hostname will not be counted as supporting HTTPS at all.

xtine:
Probably don't want to include `static/js/vendor/DataTables-1.10.6/media/images/Sorting icons.psd` as one of the files?

static/scss/main.scss
```diff
@@ -22,6 +22,8 @@ $xlarge:      new-breakpoint(max-width 1200px, 12);
 $dark-blue:       darken(#0066CC, 20%);
 $medium-blue:     darken(#0066CC, 5%);
 $light-blue:      #0066CC;
+$reward:          #2da52d;
```
xtine: As cute as this is, I would prefer we stick to variable convention with $green and $red and just leave reward/punishment as comments.

konklone:
> Probably don't want to include static/js/vendor/DataTables-1.10.6/media/images/Sorting icons.psd as one of the files?

@xtine This PR removes that file from the repo. It was there already. (The DataTables directories got moved up into `static/` instead of `static/js/vendor`.)

> As cute as this is, I would prefer we stick to variable convention with $green and $red and just leave reward/punishment as comments.

Good call, I just added a commit renaming them to `$green` and `$red`.

static/scss/includes/datatables.scss
```diff
@@ -92,6 +92,16 @@ table caption {
   table.dataTable {
     padding: 20px 0;
 
+    // reward!
+    .great {
```
xtine: Can this also be renamed to `.success` or something similar? It actually took me a few seconds to realize in the JS that great in this context doesn't mean bolder or larger.

static/js/https/domains.js
xtine: I really like this cleanup so we don't have to have this show/hide/keyboard click hack anymore.

xtine:
@konklone 
> This PR removes that file from the repo. It was there already
Oops, didn't realize it was a removal. Good stuff!

static/scss/includes/datatables.scss
konklone: @xtine Just renamed it in a subsequent commit.

cherrypj:
This change seems to have made the HTTPS table much faster, too. Much appreciated.

PR #527, opened 2016-08-28
Document the HSTS/preload field split

konklone:
Update the documentation at `/https/guidance/` to reflect the changes made in #517, where we split out HSTS and preloading into two separate fields and columns.


konklone:
Self-merging as it is a docs-only change.

PR #529, opened 2016-08-28
Fix sort order on HTTPS table

konklone:
The HTTPS columns were getting sorted with the `0` values before the `-1` values, for reasons I don't totally understand. However, converting all values to strings before sorting fixes this, and makes more conceptual sense besides.


konklone:
I added a commit that also fixes the sort order on downloadable CSVs, so that domains are in a consistent order.

PR #530, opened 2016-08-28
Sync staging branch back up with production

konklone:
This merges all the work I've done on the production site back into the staging branch, so that things don't diverge too wildly. This includes all work from the last week, including the update to DataTables and its Responsive plugin (which applies this to the accessibility tables). 

I resolved some minor merge conflicts along the way, and validated that tables all still work fine.


konklone:
Merging so as to proceed with the `staging` -> `production` cross-merge, to get the branches both back in sync with all recent changes to each other.

PR #550, opened 2016-09-15
temp data fixes

gbinal:
removing domains and agencies that are in other branches
adding back in agency info


gbinal:
this is ready to merge and push to staging

PR #573, opened 2016-09-22
Fix landing page links to not use canonical URL for permalink

konklone:
The error buttons were using the canonical URL instead of the domain to form permalinks, so it was going to `/accessibility/domain/www.data.gov` instead of `/accessibility/domain/data.gov`, which caused it to not be able to look up the data by domain in `a11y.json`.

This fixes the error buttons to always use the base domain. The canonical URL is only useful when forming links to the website itself. (Which clicking the domain link itself on the table does do.)

### DIFF
--- a/static/js/utils.js
+++ b/static/js/utils.js
@@ -54,7 +54,7 @@ var Utils = {
 
     $.each(data, function(key, value) {
       if (value) {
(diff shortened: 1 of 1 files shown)


konklone:
Self-merging for time constraints.

PR #574, opened 2016-09-22
adding a legislative agencies to the list

gbinal:
@konklone - take a glance and holler if I'm missing anything.


konklone:
You're not missing anything. It looks like between the 2016-06-30 and 2016-09-09 updates, they broke out the `Legislative Branch (Congress)` field items more granularly. In 2016-06-30, AOC was listed as `Legislative Branch (Congress)`, but is now `Architect of the Capitol`. That's a good thing, but we should check to see if there are others that changed.

PR #581, opened 2016-10-11
data updates

gbinal:


gbinal:
@konklone - this is ready to go whenever you've a moment.

PR #582, opened 2016-10-13
Fix typo in GitHub link for pshtt

siccovansas:
The URL was `https://github.com/dhs-ncat/pshtt` but this missed an `s` and should've been `https://github.com/dhs-ncats/pshtt`.


alex:
👍  lgtm

konklone:
Good catch, thank you!

PR #583, opened 2016-10-14
Addressing 'Agency' vs 'agency' bug

gbinal:
The scripts that generate the a11y agencies.json file result in an `agency` field, but the site breaks if the field isn't edited to be `Agency` instead.  I've gotten around this by doing a simple find and replace to the data before it's updated but it'll be better for it to not be an issue.  

This PR makes the site happy with a lowercase `agency` field name and reverts the data to its original form.

### DIFF
--- a/static/data/tables/accessibility/agencies.json
+++ b/static/data/tables/accessibility/agencies.json
@@ -6,7 +6,7 @@
       "Color Contrast Errors": 0,
       "Alt Tag Errors": 0,
       "pages_count": 1,
-      "Agency": "National Capital Planning Commission",
+      "agency": "National Capital Planning Commission",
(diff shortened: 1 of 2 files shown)


gbinal:
This works for me locally and should be good to merge into staging but there's no urgency and we'll need to keep an eye on the Agencies page when we do to make sure all goes well.  

Let's hold on merging this until just before the next data or site update.

konklone:
Cool, just poke again when you think this PR should be reviewed for merge, @gbinal.

konklone:
@gbinal There's a merge conflict now. Mind figuring out if it should be resolved or closed?

gbinal:
Closing in favor of https://github.com/18F/pulse/pull/668

PR #586, opened 2016-10-31
Make Pulse deployable to cloud.gov

konklone:
The site can now be run on cloud.gov:

* Production: https://pulse.fr.cloud.gov/https/domains/
* Staging: https://pulse-staging.fr.cloud.gov/https/domains/

We can update the DNS, in collaboration with the Cloud.gov team and the OMB MAX team, any time.

Note that **the data will not update automatically anymore**. We will have to manually kick off scans each week, and re-freeze the updated data into the repo and redeploy. The right place for running scans long-term in our environment is not clear, though I have copied the current scanning workhorse server AMI into our new AWS sandbox environment, and I expect to use that server for scanning for the time being.

This PR does a few things:

* Serves the app using `waitress` in production, if `PULSE_ENV` is set to `production`.
* Adds a cloud.gov manifest with a production and staging instance, both using the Python buildpack and a Python runtime of `3.5.2`.
* Causes the `/accessibility/` URLs to return a 404 based on the set environment variables, rather than the code itself. This means we don't have to maintain a weird code split between branches anymore, which will simplify future feature work considerably.
* Since we don't have to have `master` diverge from `production` anymore, this merges in all commits from the `master` branch that aren't in `production`.
* In order to freeze the data effectively into the repository and create consistent deployments, `data/db.json` is now committed to version control. It's not pretty when diffed, and even more annoyingly, it seems to change just by _running_ the app locally. But this is by far the easiest way to freeze the data into the repository, and necessitates no other architectural changes.


konklone:
You can see the env-var-driven hiding of accessibility endpoints in the cloud.gov deployment of the production and staging apps here:

* Production (404): https://pulse.fr.cloud.gov/accessibility/domains/
* Staging (200): https://pulse-staging.fr.cloud.gov/accessibility/domains/

konklone:
Since the merge of `master` makes the diff difficult, here are the commits that are new and unique to this pull request:

* Getting the app to run on cloud.gov: https://github.com/18F/pulse/pull/586/commits/d963dd22edbf6492bdbcd93b3cb56a76cc8d422f
* Adding the database to source control: https://github.com/18F/pulse/pull/586/commits/7324110c1544f10b422b76221108bf49352b8971
* Defining a staging app for cloud.gov and hiding a11y endpoints: https://github.com/18F/pulse/pull/586/commits/f4cfe9b11be66f22116f6d306ba54e06e60c4bb9

PR #588, opened 2016-11-02
Rename accessibility URLs to a11y

konklone:
Reboots the URLs as `/a11y/[whatever]` instead of `/accessibility/[whatever]`, as requested.


gbinal:
Looks great.  Thank you.  

Checks all look good.  Let's merge and push.

PR #589, opened 2016-11-02
More completely rename a11y URLs

konklone:
This updates links from the old to the new URLs.


gbinal:
Looks good.  Thank you.

PR #595, opened 2016-11-15
updating a11y and agencies

gbinal:


gbinal:
Alright - @konklone, please go ahead and merge and push this to staging in the morning.  

I hit an issue that I wasn't able to solve at all today and as a result, I'm keeping the domains.json file frozen.  There was little change in data so it shouldn't be much of an issue until we can resolve it properly with a developer.  For now though, we should go ahead and merge and deploy to staging, though.

PR #602, opened 2016-11-29
domains.json refresh from 11-14

gbinal:
This addresses some of the numbers on the domains page that hadn’t caught up from the 11-14 push.


gbinal:
@konklone - this is ready for merging and pushing to staging whenever you can.

PR #611, opened 2016-12-05
More cloud.gov production deployment work

konklone:
This tracks some further work I'm doing at updating Pulse's production cloud.gov deployment workflow and configuration.


Makefile
```diff
@@ -40,6 +46,10 @@ update_development:
 	python -m data.update --scan=skip
 
 # downloads latest snapshot of data locally
+# Pending cloud.gov production bucket:
+# cg-4adefb86-dadb-4ecf-be3e-f1c7b4f6d084
+# Pending cloud.gov backup bucket:
+# cg-72ce4caf-d81b-4771-9b96-3624b5554587
```
konklone: Credentials for these buckets can be found in the environment variables associated with the `gsa-ogp-pulse`/`pulse`/`pulse` app and `gsa-ogp-pulse`/`backups`/`s3-blank-app` app, respectively.

I'm currently taking the contents of our legacy AWS S3 bucket (bucket name `pulse.cio.gov`) and copying their full contents into each of the two buckets above, so that they are preserved.

data/update.py
```diff
@@ -61,7 +61,7 @@
 GATHER_ANALYTICS_URL = META["data"]["analytics_subdomains_url"]
 GATHER_PARENTS = DOMAINS  # Limit subdomains to set of base domains.
 GATHERERS = [
-  ["censys"],
+  ["censys", "--export"],
```
konklone: This is using the new Censys Export API feature added in https://github.com/18F/domain-scan/pull/100.

manifest.yml
```diff
@@ -2,13 +2,9 @@
 applications:
 - name: pulse
   memory: 512M
-  instances: 1
+  instances: 2
```
konklone: At cloud.gov's request, bumping instance count to 2.

manifest.yml
```diff
@@ -2,13 +2,9 @@
...
   buildpack: python_buildpack
 - name: pulse-staging
-  memory: 512M
-  instances: 1
-  buildpack: python_buildpack
-- name: pulse-subdomains
-  memory: 512M
+  memory: 128M
```
konklone: Diff here is confusing, but it's deleting the `pulse-subdomains` app (not needed, it's in production now), and dropping the `pulse-staging` memory down to 128M (it uses way less than that).

konklone:
I'm removing the `production` branch, as it is no longer used for legacy or cloud.gov deploys.

PR #621, opened 2016-12-13
Fix link to a11y section on 'About Pulse'

anand-bhat:
Link to "What is Accessibility, and why does Pulse measure it?" uses #why-a11y but section id is defined as #why-ally (lower case L). Changing to why-a11y.

### DIFF
--- a/templates/about.html
+++ b/templates/about.html
@@ -98,7 +98,7 @@
 				</p>
 			</li>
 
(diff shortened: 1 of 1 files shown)


gbinal:
Thanks for catching this.  It's been addressed by https://github.com/18F/pulse/pull/677/files.

PR #625, opened 2016-12-20
Mention securethenews in the README

konklone:
Shares some infrastructure, directly inspired, seems worth including next to the forks.


gbinal:
👍 

Thanks, @konklone.  Merging.

PR #626, opened 2016-12-21
Update dependencies to newest versions

micahsaul:
Probably not strictly necessary, but it didn't seem to break anything
and it'll make Gemnasium happy.


konklone:
:+1:

PR #632, opened 2017-01-06
Codeclimate config

micahsaul:
Removes third party packages from the Code Climate score. This should clean up our score quite a bit.

I'm leaving in the complexity and duplication checks, because eventually it might be nice to go back and refactor to fix these, but they're not pressing at all.


gbinal:
👍  - looks good!  I'll go ahead and merge.

PR #651, opened 2017-02-12
Standalone scanner, plus auto-deploy

konklone:
I've created a new scanning server instance, and set it up to perform a similar scanning process to the previous server. However, this server is not expected to also serve up one or more websites, and does not use Fabric or run nginx or other web server instances. Cloud.gov now serves pulse.cio.gov and https.cio.gov.

The new standalone server is not meant to be a long-term solution, but enough of a patch to restore automated Pulse scanning, and to avoid requiring a Pulse team member to perform manual deploys to update the production website. The long-term solution should still be to find a cloud.gov solution for the (extensive) scanning process, and to decouple the scanning process from depending on persistent disk.

I've left inline comments on the changed files to explain individual pieces. The main things to understand are that:

* There is now a `scan-box-deployer` service instance in the `gsa-ogp-pulse`/`pulse` org/space. The process for creating this service and generating deploy-ready credentials is identical to the process [documented for Travis](https://cloud.gov/docs/apps/continuous-deployment/), but we do not use Travis to perform the deployment. Instead, the scanning server has the relevant credentials stored as environment variables in `$HOME/.bashrc`, which is `source`'d during the scanning process.
* The `pulse.cio.gov` S3 bucket, which is meant to be an ongoing archive for Pulse data, has been abandoned in favor of a cloud.gov-managed bucket. I did populate the cloud.gov-managed bucket with the contents of the `pulse.cio.gov` bucket at the beginning of December 2016, but there are several scans which haven't yet been copied from the old one to the new one. There is also a cloud.gov backup bucket, tied to `gsa-ogp-pulse`/`backups`. That also saw an initial population in December 2016, but has not seen any further population. It should be caught up, and then probably also auto-populated from the main bucket on a regular interval. I'm generally not satisfied with the level of backups we're performing for this data. If Pulse historical scan data is lost, it cannot be recreated.


static/js/https/domains.js
```diff
@@ -137,7 +137,7 @@ $(document).ready(function () {
   };
 
   var agencyDownloadFor = function(row) {
-    return "https://s3.amazonaws.com/pulse.cio.gov/live/subdomains/agencies/" + row["agency_slug"] + "/https.csv";
+    return "https://s3-us-gov-west-1.amazonaws.com/cg-4adefb86-dadb-4ecf-be3e-f1c7b4f6d084/live/subdomains/agencies/" + row["agency_slug"] + "/https.csv";
```
konklone: It would be good to use the value in `meta.yml` (or an env var or whatever) instead of hardcoding the bucket URL, but this requires some kind of refactoring to achieve, since this JS file is served statically instead of through a Flask view.

templates/data.html
```diff
@@ -30,7 +30,7 @@
 			<h4>Site Wide</h4>
 			<p>Datasets that power the entire pulse.cio.gov website.</p>
 				<ul><li>
-				Domain Inspect Results - <a href="https://s3.amazonaws.com/pulse.cio.gov/live/scan/inspect.csv">CSV</a>
+				Domain Inspect Results - <a href="https://s3-us-gov-west-1.amazonaws.com/cg-4adefb86-dadb-4ecf-be3e-f1c7b4f6d084/live/scan/inspect.csv">CSV</a>
```
konklone: It would be better to use `meta.yml`, which should be possible in the template -- I just didn't bother doing the refactor here yet.

konklone:
Please hold on merging until I can verify that a fully cron-initiated scan, using a copy of the code checked out from this PR's branch, completes and uploads and deploys successfully. Marking in the title for emphasis, and I will update here when the result is verified, which should be in < 2 days.

smarina04:
Thanks so much, Eric. While I realize that this is not long-term, it's a
huge step forward. (Although I must confess I was somewhat excited about
learning how to initiate those scans myself if Gray was not available :)

Kidding aside, just to make sure I'm clear on the back-up issue. Who at
18f will be working on making sure that Pulse data is properly backed up
for historical analysis? Micah? Not being to recover this data puts us at a
huge risk, especially as we expand our services/modules.

Thanks again,
Marina

konklone:
Worked perfectly on the first try! Scan fully completed and auto-deployed, data is now updated as of February 12th:

https://pulse.cio.gov/https/domains/

Latest scan data is also uploaded successfully to S3:

https://s3-us-gov-west-1.amazonaws.com/cg-4adefb86-dadb-4ecf-be3e-f1c7b4f6d084/live/scan/meta.json

This is ready for merging.

data/update.py
```diff
@@ -204,14 +204,10 @@ def gather_subdomains(gatherer, command):
       "--output=%s" % gatherer_output,
       "--parents=%s" % GATHER_PARENTS,
       "--sort",
+      "--force",
```
konklone: Noticed a bug while doing this -- by leaving off the `--force` flag during gathering, this was reusing cached data from Censys.io. This means our Censys hostname seed data was stale for a while. (The scans of each domain's current compliance status was properly checked each scan.)

PR #655, opened 2017-02-16
Added extra id attributes to allow deeper linking

IanLee1521:
Was trying to link down specifically to the definition for "Preloaded", and noticed that some of the ``id`` attributes were missing in the guide.

### DIFF
--- a/templates/https/guide.html
+++ b/templates/https/guide.html
@@ -62,7 +62,7 @@
 
       <article>
 
-        <h4>Implementing HTTPS</h4>
+        <h4 id="implementing">Implementing HTTPS</h4>
 
         <p>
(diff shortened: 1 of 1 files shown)


konklone:
Thank you! :+1:

konklone:
@IanLee1521 Any time sensitivity to deploying this (i.e. are you making any reports or sites that make use of this)? We don't have auto-deploys hooked up right now from GitHub, so it needs a staff member to deploy, and I'm not able to do that for another week or so. If it's time-sensitive, I can ask a colleague to run a deploy.

IanLee1521:
That's fine. Not overly time sensitive. I worked around it for now.

PR #658, opened 2017-03-01
Updating ineligible DAP domains

tdlowden:
FYI, hud.gov and america.gov have no 2nd level domain endpoint. They redirect to subdomains. Not a best practice, but it is impossible for them to have DAP code on those hostnames, regardless.


gbinal:
Looks good.  Let's revisit the scenario around america.gov but for now, I'll go ahead and merge. Thanks, @tdlowden!

Issue #660, opened 2017-03-03
DHS Domains to be removed from DAP


mtharmon20:
Morning -

The following DHS domains do not have publicly accessible landing pages and DHS has determined they are not eligible for DAP due to privacy and security reasons:

cpnireporting.gov
nmsc.gov

Can these please be removed from the Pulse dashboard?  Thanks!

Matt

tdlowden:
Added these to the "ineligible" list for PR #662

gbinal:
I've merged #662.  These changes should take effect in the next scan (early next week).

mtharmon20:
Any update on this - these two domains are still appearing in the DAP listing for DHS.

tdlowden:
Hey all. It seems the scan has not taken the updated "not eligible" list into account. Can we revisit?

konklone:
@tdlowden @gbinal I think it's because this hasn't been merged/deployed yet: https://github.com/18F/pulse/pull/661

I'm merging and deploying it now, and updating the scan server with the latest code directly. That should update the ineligible domains for the next scan.

PR #661, opened 2017-03-07
Update oneself during scan-deploy process

konklone:
While we still have the scanning living on a persistent box, make sure its auto-deploy process is always taking into account the latest code. This runs a `git pull` before _and_ after the ~40 hour scanning process.


konklone:
This is now deployed to the scanner, and should resolve issues like #660 on the next scan.

PR #662, opened 2017-03-07
Update analytics.yml

tdlowden:
Alphabetized and added cpnireporting.gov and nmsc.gov per issue #660


gbinal:
looks good.  merging.

Issue #664, opened 2017-03-10
Integrate a new sites.csv file from analytics.usa.gov


gbinal:
The analytics.usa.gov project wanted to add a slightly higher threshold for the [generation](https://github.com/18F/analytics-reporter/blob/master/reports/usa.json#L352-L370) of it's [sites.csv report](https://analytics.usa.gov/data/live/sites.csv) (which provides a list of participating sites).  This was achieved with https://github.com/18F/analytics.usa.gov/issues/449.  

We didn't realize at the time though that pulse [uses that file](https://github.com/18F/pulse/blob/master/meta.yml#L19) as a means of discovering more subdomains for the https scans and that it would prefer the threshold to stay low.  

To resolve this, we're going to generate a new report that uses the lower threshold and update the url that pulse looks to to point to this new file instead.

gbinal:
I've created https://github.com/18F/analytics-reporter/pull/203 to create a new report called `sites_extended` that'll live [here](https://analytics.usa.gov/data/live/sites_extended.csv).  

When it's merged and pushed to production, we should review, merge, and push to production  https://github.com/18F/pulse/pull/665

konklone:
This is integrated.

PR #665, opened 2017-03-10
changing analytics_subdomains_url link

gbinal:
to help address https://github.com/18F/pulse/issues/664

### DIFF
--- a/meta.yml
+++ b/meta.yml
@@ -16,6 +16,6 @@ data:
   analytics_url: "https://analytics.usa.gov/data/live/second-level-domains.csv"
 
   # DAP URL where we source public .gov subdomains from.
-  analytics_subdomains_url: "https://analytics.usa.gov/data/live/sites_extended.csv"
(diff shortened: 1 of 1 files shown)


meta.yml
```diff
@@ -16,6 +16,6 @@ data:
   analytics_url: "https://analytics.usa.gov/data/live/second-level-domains.csv"
 
   # DAP URL where we source public .gov subdomains from.
-  analytics_subdomains_url: "https://analytics.usa.gov/data/live/sites.csv"
+  analytics_subdomains_url: "https://analytics.usa.gov/data/live/sites-extended.csv"
```
konklone: We'll need to also update the deploy script in analytics-reporter to generate and publish this report.

konklone: Actually, I don't think we will. But I won't merge this until the URL works.

konklone: It works! https://analytics.usa.gov/data/live/sites-extended.csv

Issue #674, opened 2017-03-24
Remove Accessibility Verbiage from "About"


smarina04:
Remove Accessibility tab and verbiage from the production "About" page on Pulse since the Module is still in staging.

gbinal:
https://github.com/18F/pulse/pull/677 should address this.

gbinal:
done

PR #677, opened 2017-03-24
removing a11y language

gbinal:


gbinal:
This is to address #674.

gbinal:
Thanks, @micahsaul - Is this pushed to production yet?  If not, game to do that sometime today?

PR #680, opened 2017-04-04
adding another to the DAP exclusion list

gbinal:
@tdlowden - this would add `nationalbank.gov` to the list since it redirects to https://www.banknet.gov/entrance/default.html.


gbinal:
This checks out, so we  should be good to merge.

Issue #689, opened 2017-04-27
update the views.py to ensure that customer satisfaction section only pushes to staging 


gbinal:
The customer satisfaction entries in the [views.py](https://github.com/18F/pulse/blob/adding-custsat-section/app/views.py) should be modified like the a11y section so that that section doesn't push to production.

micahsaul:
This was addressed in #690

PR #690, opened 2017-04-27
[WIP] Adding custsat section

gbinal:
We should review this PR to ensure that it doesn't endanger any parts of the production site.  Then, after https://github.com/18F/pulse/issues/689, we should go ahead and merge and push this branch to staging so that we can continue refining it online instead of just locally.


gbinal:
@micahsaul - To confirm, I've reviewed this PR and all that it does is add files for the customer satisfaction section and make a few additions to the views.py file.  That last part is the only thing that could endanger the rest of the site and I've given it a good look over, so I think we're okay to merge this and push it whenever we're ready.  

We'll of course want to keep an eye on the site in case we have to roll back after.

gbinal:
Let's merge this.

Issue #650, opened 2017-02-08
Domains to be removed from DAP compliance list


gsolomon123:
Hello,
The following DOJ domains should be removed from DAP. 
Internal training sites: 
Learn ATF
Learn DOJ
cjis.gov

Redirect:
https://www.medalofvalor.gov

konklone:
Hi @gsolomon123 - can you be more specific about which domains you mean?

For medalofvalor.gov, we do catch and remove server-side redirects, but medalofvalor.gov uses a meta redirect, which we do not yet catch. If you update your server to redirect using a 3XX status code on the server, it'll cause the site to automatically disappear from Pulse.

Also, separately, the HSTS header for medalofvalor.gov is incorrect -- it uses `always` instead of `preload`:

```
$ curl --head https://www.medalofvalor.gov

HTTP/1.1 200 OK
...
Strict-Transport-Security: max-age=31536000; includeSubDomains; always
```

There is no `always` directive for HSTS that I am aware of. I think you meant to put `preload` there. I'm looking forward to seeing `medalofvalor.gov` preloaded!

gsolomon123:
Hi Eric,
My question about Medalofvalor.gov was regarding DAP. I was of the understanding that redirects are not required to have the DAP code. The websites I referenced in my submission to GitHub (which is what I believe you’re responding to) were all with regard to DAP code missing from the header.

I asked for these domains to be removed from the Analytics module of Pulse because they are for internal training and are password protected sites.

Learn ATF
Learn DOJ
cjis.gov
I’m not sure why a redirect would be required to contain the DAP code.

Thank you,
Gail

Gail D. Solomon
DOJ | JMD | Office of the Chief Information Officer
Content Manager, Justice.gov<https://justice.gov/>, Collaboration and Web Services
gail.d.solomon@usdoj.gov<mailto:gail.d.solomon@usdoj.gov>
Office: (202) 307-0589
Cell: (202) 598-3597

If not available please contact Cathy Flanagan<mailto:cathy.flanagan@usdoj.gov>.

konklone:
@gsolomon123 Please re-read my previous message. The issue is in the _kind_ of redirect you're employing. If you move to a server-side redirect, Pulse will automatically remove such domains from eligibility.

We are looking into detecting meta redirects as well, which will also fix this problem, but for quickest results, you can update your servers to use 301 redirects.

konklone:
[...]

PR #704, opened 2017-05-15
Update analytics.yml

tdlowden:
Adding DOJ internal training sites as per https://github.com/18F/pulse/issues/650.

These sites do have public landing pages, but they are intended solely for DOJ employees, so I am removing them from the list.


gbinal:
Roger.  Tests look good and I've given it an 👀 .  Merging now.

PR #711, opened 2017-06-21
Move a11y into the Pulse processing pipeline

micahsaul:
This adds the processing of a11y data from the domain-scan repo to the Pulse pipeline. Things it does not do (to come in follow-up PRs):

1) Actually turn on the automatic scanning of a11y data. There are some unresolved issues in the a11y scan that make me want to hold back for now.

2) Switch the presentation layer to use the newly processed data. This is a trivial task, but since the scan is not automatically running yet, it is premature.

3) Generate the data/domains/a11y.csv file. Same reason as above.

Next step is to pull the a11y processing stuff out of domain-scan and clear up the last scanning issues before switching it on. PRs in that repo to come.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -200,7 +200,7 @@ class Agency:
   # An agency which had at least 1 eligible domain.
   def eligible(report_name):
     return db.table('agencies').search(
-      (Query()[report_name]['eligible'] > 0)
+      Query()[report_name]['eligible'] > 0
     )
 
   # Create a new Agency record with a given name, slug, and total domain count.
--- a/app/models.py
+++ b/app/models.py
@@ -200,8 +200,7 @@ class Agency:
   # An agency which had at least 1 eligible domain.
   def eligible(report_name):
     return db.table('agencies').search(
-      (Query()[report_name]['eligible'] > 0) |
-      (Query()[report_name]['pages_count'] > 0)
+      (Query()[report_name]['eligible'] > 0)
     )
 
   # Create a new Agency record with a given name, slug, and total domain count.
--- a/app/models.py
+++ b/app/models.py
@@ -200,7 +200,8 @@ class Agency:
   # An agency which had at least 1 eligible domain.
   def eligible(report_name):
     return db.table('agencies').search(
-      Query()[report_name]['eligible'] > 0
+      (Query()[report_name]['eligible'] > 0) |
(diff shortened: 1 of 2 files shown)


gbinal:
Looks good to me, @micahsaul.  @konklone ,- if you're game to take a look, it'll be obliged.   Otherwise, perhaps we merge and deploy tomorrow morning PT, @micahsaul and carefully monitor afterwards.

app/models.py
```diff
@@ -200,7 +200,8 @@ class Agency:
   # An agency which had at least 1 eligible domain.
   def eligible(report_name):
     return db.table('agencies').search(
-      Query()[report_name]['eligible'] > 0
+      (Query()[report_name]['eligible'] > 0) |
+      (Query()[report_name]['pages_count'] > 0)
```
konklone: Why not have the `pages_count` determination code determine the value for `eligible`?

data/processing.py
```diff
@@ -712,15 +806,21 @@ def latest_reports():
...
 
 # Hacky helper - print out the %'s after the command finishes.
 def print_report():
   print()
 
   report = Report.latest()
   for report_type in report.keys():
-    if report_type == "report_date":
+    # The a11y report has a very different use than the others
+    if report_type == "report_date" or report_type == "a11y":
```
konklone: Could you help me understand how the `a11y` report differs from the others?

micahsaul: The a11y report is basically just a full dump of the scan data, with line by line errors per domain. The others are high level summaries.

app/models.py
micahsaul: Heh, because I wasn't thinking. The previous process_a11y script didn't use "eligible" and I was keeping things the same to make sure that if folks were using the previous report, things wouldn't break. But there's probably no reason not to just add "eligible" to the report as well, containing the same data.

konklone: Got it -- thanks for addressing this comment in subsequent commits.

konklone:
@micahsaul Fine from me to merge. :+1: Are you still planning to push anything here, or is this good to go?

micahsaul:
@konklone nope, this is set. I'll have a follow up PR to switch the site to the new data once it's being consistently run. Thanks!

micahsaul:
Woot! Thanks @konklone!

PR #717, opened 2017-08-01
Move .gov specific a11y config from domain scan

micahsaul:


gbinal:
Looks good to me.

konklone:
Thank you!

PR #724, opened 2017-08-23
Run a11y and cust-sat on the first of the month

micahsaul:


deploy/crontab
```diff
@@ -3,3 +3,4 @@
 # Currently, they take ~40 hours with single-threaded SSL Labs API requests
 # to 500+ domains, and subdomain gathering and scanning.
 0 0 * * 1,3,5 PULSE_ENV=production /bin/bash /opt/scan/pulse/deploy/cron.sh > /opt/logs/pulse/cron.log 2>&1
+0 0 1 * * SCANNERS=pshtt,a11y,third_parties /bin/bash /opt/scan/pulse/deploy/cron.sh > /opt/logs/pulse/cron.log 2>&1
```
micahsaul: Not sure if PULSE_ENV is needed here since it's also set in cron.sh

micahsaul: @konklone, thoughts on this? Do you recall why you specifically set PULSE_ENV both here and in cron.sh?

konklone: Could just be redundant, or predating the existence of that line in cron.sh.

konklone:
To note though, the PR doesn't set it for the first of the month, but rather includes it in the every 2-3 days scan.

@micahsaul, have you benchmarked how long it takes to run the `a11y` and `third_parties` scanners on the list of federal parent .gov domains?

PR #726, opened 2017-08-28
Add third-party and a11y to normal scan

micahsaul:
I reverted the previous PR because I forgot the a11y config files. These
two scans take ~45 minutes combined, so this _shouldn't_ impact things
too much. Once we've confirmed that the data looks good, we'll switch
back to the normal scan and only run a11y and customer satisfaction on
the first of the month.


gbinal:
Sounds good to me.  Let's go ahead and merge.

PR #728, opened 2017-09-11
Fix scan name

micahsaul:
Pronouncing it "ally" in my head as opposed to "a-eleven-y" has lead to more typos in this app than anything else.


gbinal:
👍

PR #736, opened 2017-10-31
Revamp HTTPS table and refactor to better support subdomains

konklone:
This PR is a major refactor of the Pulse data backend, import/processing pipeline, and HTTPS display system. The major goals are:

* Have the top-line HTTPS reporting show a % of all known services, not parent domains.
* Have the top-line HTTPS reporting show the % complying with federal policy, not just "Uses HTTPS".
* Support BOD 18-01 ([cyber.dhs.gov](https://cyber.dhs.gov)) by adding a column measuring support for RC4/3DES/SSLv2/SSLv3, and including BOD 18-01 compliance in the top line measurement.
* Support subdomains in a native, non-confusing way, and provide a clear (and performant) way to see all of them and download them to CSV.
* Replace SSL Labs with SSLyze for TLS/SSL protocol-level analysis.
* Eliminate confusion in agencies by fully matching the presentation of [DHS HTTPS reporting](https://18f.gsa.gov/2017/01/06/open-source-collaboration-across-agencies-to-improve-https-deployment/) so that we are showing the same numbers and data, aggregated in the same way.

screenshot from 2017-10-31 17-10-37

screenshot from 2017-10-31 17-10-54

To do so, this makes significant internal adjustments:

* Adjusts the scanners and gatherers to use the new [multi-gatherer](https://github.com/18F/domain-scan/pull/124) in domain-scan, which consolidates gather results from multiple sources into a single CSV file that dedupes hostnames and notes which sources each hostname appears in. 
* Refactors `processing.py` to make use of the consolidated subdomain data, and to load every subdomain into the `Domain` table.
* Refactors the `Domain` table to hold both parent domains and subdomains, and refactors the filtering and exporting code to account for this.
* Redesigns the HTTPS table to drop the `Uses` column, to add a BOD 18-01 crypto column, and to add a full compliance column. (The table shows one column for BOD 18-01, the CSV also breaks it out into individual columns.)
* Rewires the HTTPS table to show data for each subdomain, inline, when expanded per-parent domain. This is meant to be helpful for both usability and performance.
* CSVs can now be downloaded for the entire HTTPS table (including all subdomains) or for all subdomains of a particular parent domain. CSVs are linked within the table.
* Gzip compression is added, because OBVIOUSLY, but also because there's now more data we're transferring due to increased subdomain data stored and served.
* Updates the HTTPS guidance page to document all the changes and link to additional information.


data/processing.py
```diff
@@ -345,185 +376,348 @@ def load_scan_data(domains):
...
+  return full
 
 
-# TODO: A domain can also be eligible if it has eligible subdomains.
-#       Has display ramifications.
 def eligible_for_https(domain):
-  return (domain["live"] == True)
+  return (
+    (domain["live"] == True) and
+    (domain["branch"] == "executive")
```
konklone: In theory, we could drop the `domain["branch"] == "executive"` check from the eligibility check, but it doesn't hurt to be explicit.

data/update.py
```diff
@@ -141,41 +117,40 @@ def run(options):
...
-  live_db = "s3://%s/live/db/" % (BUCKET_NAME)
+  live_parents = "s3://%s/live/parents/" % (BUCKET_NAME)
   live_subdomains = "s3://%s/live/subdomains/" % (BUCKET_NAME)
-  archive_scanned = "s3://%s/archive/%s/scan/" % (BUCKET_NAME, date)
-  archive_cached = "s3://%s/archive/%s/cache/" % (BUCKET_NAME, date)
-  archive_db = "s3://%s/archive/%s/db/" % (BUCKET_NAME, date)
+  live_db = "s3://%s/live/db/" % (BUCKET_NAME)
   archive_subdomains = "s3://%s/archive/%s/subdomains/" % (BUCKET_NAME, date)
+  archive_parents = "s3://%s/archive/%s/parents/" % (BUCKET_NAME, date)
+  archive_db = "s3://%s/archive/%s/db/" % (BUCKET_NAME, date)
```
konklone: This S3 uploading code is not fully tested yet, but given the weight of this PR, I think it's okay to merge this as-is and then we can file a new PR fixing any issues here.

data/update.py
```diff
@@ -141,41 +117,40 @@ def run(options):
...
   a11y_redirects = "--a11y_redirects=%s" % A11Y_REDIRECTS
   a11y_config = "--a11y_config=%s" % A11Y_CONFIG
 
   full_command =[
     SCAN_COMMAND, DOMAINS,
-    scanners, analytics, output, a11y_config, a11y_redirects,
+    scanners,
+    analytics, a11y_config, a11y_redirects,
+    output,
+    "--sslyze-no-certs", # ugh, temporary
```
konklone: Noting that this will change soon in a release of `domain-scan`, probably to become `--sslyze-certs=false`. Not a blocker.

data/update.py
```diff
@@ -185,69 +160,66 @@ def scan(options):
...
+  full_command += GATHERER_OPTIONS
 
-    full_command = [GATHER_COMMAND]
-    full_command += command
+  # Common to all gatherers.
+  # --parents gets auto-included as its own gatherer source.
+  full_command += [
+    "--output=%s" % SUBDOMAIN_DATA_GATHERED,
+    "--suffix=%s" % GATHER_SUFFIX,
+    "--ignore-www",
```
konklone: `--ignore-www` is a new flag in domain-scan which causes the gatherer to collapse `www.history.example.gov` and `history.example.gov` into the same entry.

data/update.py
```diff
@@ -185,69 +160,66 @@ def scan(options):
...
+  # In debug mode, use cached data, and allow easy Ctrl-C.
+  if options.get("debug"):
+    full_command += ["--serial"]
 
-    # In real mode, ignore cached data, and parallelize.
-    else:
-      full_command += ["--force"]
+  # In real mode, ignore cached data, and parallelize.
+  elif not options.get("cache"):
+    full_command += ["--force"]
```
konklone: Noting that in a future release of domain-scan, `--force` will become the default, and `--cache` will be needed to opt-in to making use of caching (which is only when `--debug` or `--cache` is passed to this script).

manifest.yml
```diff
@@ -9,7 +9,7 @@ applications:
     NEW_RELIC_APP_NAME: Pulse | Prod
     NEW_RELIC_ENV: production
 - name: pulse-staging
-  memory: 128M
+  memory: 512M
```
konklone: The staging site had an out of memory error, so I bumped it up to match the production site.

meta.yml
```diff
@@ -10,12 +10,19 @@ description: "How federal government domains are meeting best practices on the w
 # Data sources. These URLs should have their content updated in-place.
 data:
   # URL where we download the .gov domain list from.
-  domains_url: "https://raw.githubusercontent.com/GSA/data/gh-pages/dotgov-domains/current-federal.csv"
+  domains_url: "https://github.com/GSA/data/raw/master/dotgov-domains/current-federal.csv"
```
konklone: The DotGov program has moved their URLs away from `gh-pages` and to `master`. See https://github.com/GSA/data/issues/86

meta.yml
```diff
@@ -10,12 +10,19 @@ description: "How federal government domains are meeting best practices on the w
...
 
   # URL where we download the participating DAP .gov list from.
   analytics_url: "https://analytics.usa.gov/data/live/second-level-domains.csv"
 
   # DAP URL where we source public .gov subdomains from.
   analytics_subdomains_url: "https://analytics.usa.gov/data/live/sites-extended.csv"
 
+  # Preprocessed CSV where we source End of Term Archive data from.
+  eot_subdomains_url: https://github.com/GSA/data/raw/master/end-of-term-archive-csv/eot-2016-seeds.csv
```
konklone: We weren't previously including EOT 2016 data. Now we are, which significantly increases the scope of the subdomains we're scanning.

static/js/accessibility/agencies.js
```diff
@@ -1,59 +1,20 @@
...
-    var table = $("table").DataTable({
-      responsive: true,
-      initComplete: Utils.searchLinks,
-
-      data: data,
+  // break caching
+  var time = (new Date()).getTime();
 
+  $.get("/static/data/tables/accessibility/agencies.json?" + time, function(data) {
+    Tables.initAgency(data.data, {
```
konklone: The `Tables.initAgency` method is now defined centrally, and handles most common configuration of DataTables across all our pages.

static/js/accessibility/domains.js
```diff
@@ -1,58 +1,27 @@
...
-  var renderTable = function(data) {
-    var table = $("table").DataTable({
-
-      responsive: true,
-
-      data: data,
-
-      initComplete: Utils.searchLinks,
+  // break caching
+  var time = (new Date()).getTime();
```
konklone: For those JS files making requests to `/static/` (the `a11y` and `customer-satisfaction` pages), a cache-breaking timestamp is passed, which avoids people needing to hard refresh upon deployment.

static/js/analytics/domains.js
```diff
@@ -1,58 +1,22 @@
...
           cellType: "th",
-          render: Utils.linkDomain
+          render: Tables.canonical
         },
         {data: "canonical"},
         {data: "agency_name"},
         {
           data: "analytics.participating",
-          render: display(names.dap)
+          render: Tables.boolean
```
konklone: The common case of `False`->`No`, `True`->`Yes` is now handled by this utility function.

static/js/analytics/domains.js
```diff
@@ -1,58 +1,22 @@
...
-      initComplete: Utils.searchLinks,
+      csv: "/data/hosts/analytics.csv",
 
       columns: [
         {
           data: "domain",
           width: "210px",
           cellType: "th",
-          render: Utils.linkDomain
+          render: Tables.canonical
```
konklone: The common case of linking to a row's canonical URL is now handled by this utility function.

static/js/dataTables.lengthLinks.js
```diff
@@ -1,90 +0,0 @@
-/*! Page length control via links for DataTables
```
konklone: Length links were removed entirely, so the plugin is being yanked.

static/js/https/domains.js
```diff
@@ -1,7 +1,70 @@
-$(document).ready(function () {
+$(function () {
```
konklone: I won't annotate the whole file, but the JavaScript for rendering the domains table here has been almost completely refactored.

static/js/vendor/dataTables.lengthLinks.js
```diff
@@ -1,90 +0,0 @@
-/*! Page length control via links for DataTables
```
konklone: The `lengthLinks` plugin was in source control twice for some reason.

templates/https/agencies.html
```diff
@@ -1,8 +1,8 @@
 {% extends "layout.html" %}
```
konklone: I adjusted the strings on the HTTPS screen to assume a little more knowledge by agencies, since we're focusing more on compliance and less on general HTTPS monitoring.

templates/https/guide.html
```diff
@@ -1,8 +1,8 @@
 {% extends "layout.html" %}
```
konklone: The HTTPS guidance page was rewritten in many places to describe the changes.

templates/includes/footer.html
```diff
@@ -1,9 +1,6 @@
 <footer role="contentinfo">
 	<div class="wrapper">
 
-    Built <a href="https://github.com/18F/pulse">in the open</a> as a collaboration between <a href="https://18f.gsa.gov">18F</a> and the <a href="http://gsa.gov/governmentwidepolicy">Office of Government-wide Policy</a> at the <a href="http://gsa.gov">General Services Administration</a>.
-
-    <p>SSL Labs grades are used with permission, and powered by the <a href="https://www.ssllabs.com/">Qualys SSL Labs API</a>.
-    </p>
+    Built <a href="https://github.com/18F/pulse">in the open</a> as a collaboration between <a href="https://18f.gsa.gov">18F</a> and the <a href="https://www.gsa.gov/governmentwidepolicy">Office of Government-wide Policy</a> at the <a href="https://www.gsa.gov">General Services Administration</a>.
```
konklone: Since we're now no longer using SSL Labs, I removed the SSL Labs citation from the bottom of the site.

templates/includes/head.html
```diff
@@ -48,7 +48,7 @@
 <!-- Scripts
 ================================================== -->
 <script src="/static/js/vendor/modernizr-2.8.3.min.js"></script>
-<script src="/static/js/vendor/jquery.min.js"></script>
+<script src="/static/js/vendor/jquery-3.2.1.min.js"></script>
```
konklone: I updated jQuery to 3.2.1, and put the version number explicitly in the filename.

gbinal:
Heck yeah.  👍

data/update.py
micahsaul: Concur

static/js/accessibility/domains.js
micahsaul: Fantastic! Thank you for this!

PR #737, opened 2017-11-08
Nov a11y and cust sat

micahsaul:


gbinal:
thanks, @micahsaul.  @konklone - if it's not a problem, I'd love to go ahead and push these to staging today.

konklone:
Deployed to staging. I also added a cache-breaker for the XHR call for all 4 of these reports, so that people don't need to hard-refresh the pages to see the changes:

https://github.com/18F/pulse/commit/bbf6c09c9eb14261bb2493fffd9fc50b3bfa9b34

PR #738, opened 2017-11-10
updating scan date for the a11y section

gbinal:


gbinal:
Alas, this small detail was missing from #737.  

@micahsaul or @konklone - it's a very minor change.  Either of you mind merging and pushing or if I go ahead and do it?

PR #742, opened 2017-11-24
Deploy/S3 fixes for new scanning process

konklone:
This does two main things:

* Updates the S3 uploading/archiving process to account for the new directory structure and overall processing pipeline.

* Updates the S3 _downloading_ process (whether done via `make data_init` or via `python -m data.update --scan=download`) to also account for the new directory structure. It also moves all the logic in one place inside `make data_init`, and has it hinge on the defined scanners for parent domains and subdomains, so that there's less duplication of code and the S3 logic is defined in one place.

There's also a tweak to the production and staging memory sizes, to account for increased vuln scanning to the production site that may have been leading to increased crashes.

### DIFF
--- a/data/update.py
+++ b/data/update.py
@@ -121,14 +121,18 @@ def run(options):
 
 # Upload the scan + processed data to /live/ and /archive/ locations by date.
 def upload_s3(date):
+  # Used for all operations.
   acl = "--acl=public-read"
+
+  # Used when uploading to the live/ dir.
+  delete = "--delete"
 
   live_parents = "s3://%s/live/parents/" % BUCKET_NAME
   live_subdomains = "s3://%s/live/subdomains/" % BUCKET_NAME
   live_db = "s3://%s/live/db/" % BUCKET_NAME
 
-  shell_out(["aws", "s3", "sync", PARENTS_DATA, live_parents, acl])
-  shell_out(["aws", "s3", "sync", SUBDOMAIN_DATA, live_subdomains, acl])
+  shell_out(["aws", "s3", "sync", PARENTS_DATA, live_parents, acl, delete])
+  shell_out(["aws", "s3", "sync", SUBDOMAIN_DATA, live_subdomains, acl, delete])
   shell_out(["aws", "s3", "cp", DB_DATA, live_db, acl])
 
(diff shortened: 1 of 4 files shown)


konklone:
Tested and confirmed in production.

data/update.py
```diff
@@ -108,38 +112,83 @@ def run(options):
...
-  archive_db = "s3://%s/archive/%s/db/" % (BUCKET_NAME, date)
-
+def upload_s3(date):
+  # Used for all operations.
   acl = "--acl=public-read"
 
-  shell_out(["aws", "s3", "sync", PARENTS_DATA, live_scanned, acl])
-  shell_out(["aws", "s3", "sync", SUBDOMAIN_DATA, live_subdomains, acl])
+  # Used when uploading to the live/ dir.
+  delete = "--delete"
```
konklone: Note that the `aws s3 sync` command is now using `--delete` when uploading a new scan to S3. We weren't using this before, which meant that old data was getting copied into new scans. This wouldn't impact the results, but would mean that the `cache/` dir was growing steadily. There were `cache/inspect/` directories in the latest scan, even though we moved from `site-inspector` to `pshtt` over a year ago.

data/update.py
```diff
@@ -108,38 +112,83 @@ def run(options):
...
+  shell_out(["aws", "s3", "sync", live, archive, acl])
+
+
+# Makes use of the public URLs so that this can be run in a dev
+# environment that doesn't have write credentials to the bucket.
+def download_s3():
+
+  def download(src, dest):
+    # remote sources relative to bucket
+    url = "https://s3-us-gov-west-1.amazonaws.com/%s/%s" % (BUCKET_NAME, src)
```
konklone: Note: the AWS region is hardcoded here. This should probably be moved out to `meta.yml` or something, alongside the bucket name.

manifest.yml
```diff
@@ -1,15 +1,15 @@
 ---
 applications:
 - name: pulse
-  memory: 512M
+  memory: 768M
```
konklone: This is the max we can fit for the production site. I'm not sure how we're hitting our 2GB quota in the org, given what I can see by inspecting the contents via `cf`, but this is the ceiling.

PR #743, opened 2017-11-25
Switch to Censys snapshot, add Reverse DNS snapshot data

konklone:
This switches us to a snapshot of data from a recent Censys.io export (as we will be on a hiatus from using their service until we set up a Google Cloud account to bill BigQuery queries to), and adds a new subdomain source: [Reverse DNS data collected by Rapid7](https://scans.io/study/sonar.rdns_v2). 

The Reverse DNS data is also a snapshot, as the full dataset is a 130GB [JSON Lines](http://jsonlines.org) file that is non-trivial to parse and filter. I expect eventually we'll get more sophisticated at processing this data automatically, but for now, it's updated "whenever". It adds 2,850 unique hostnames that respond to HTTP/HTTPS and which don't appear in other sources.

### DIFF
--- a/data/env.py
+++ b/data/env.py
@@ -40,11 +40,12 @@ SCANNERS = scanner_string.split(",")
 GATHER_SUFFIXES = os.environ.get("GATHER_SUFFIXES", ".gov,.fed.us")
 
 # names and options must be in corresponding order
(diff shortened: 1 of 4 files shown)


konklone:
Superseded by #745.

PR #746, opened 2017-12-11
Turn on Lambda in production

konklone:
This enables `--lambda` mode for `pshtt` and `sslyze` scans in production. It relies on the `--lambda-profile` flag recently added to domain-scan, to use a separate AWS named profile from the one used to speak to the cloud.gov S3 bucket.

It also documents the use of the Rapid7 Reverse DNS data source for subdomains.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,8 +37,9 @@ clean:
 # Production data update process:
 #
 # Run a fresh scan, update the database, and upload data to S3.
+# Enable Lambda mode, using Lambda AWS profile set up in production.
 update_production:
-	python -m data.update --scan=here --upload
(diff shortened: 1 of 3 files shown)


konklone:
A scan that used this branch is published at:

https://pulse.cio.gov/https/domains/

However, the a11y and phantomas executables (and the `node` binary they use) are still not set up for `cron`-based execution (the path to `node` cannot be found, as the `phantomas` executable is a script referencing `/usr/bin/env node`, which is not defined in cron's environment). 

So, they do run locally as expected, even intermingled with Lambda-based scanners, but the output isn't suitable for staging, where those datasets are used. However, this has been the status quo since before this PR.

PR #750, opened 2017-12-20
Updates to allow processing to contiue without Analystics data

jeremy-gillikin:
This allows the processing portion to complete if one chooses to use only the lambda ready scanners (pshtt, sslyze).


konklone:
:+1: Looks good, thank you!

PR #753, opened 2018-01-10
Add Janurary cust-sat and a11y results

micahsaul:


gbinal:
Looks good.  Thanks, @micahsaul - game to go ahead and merge/deploy?

konklone:
@gbinal @micahsaul The nightly scan process does an automatic `git pull` before and after the scan that should deploy these changes automatically. Want to just let them appear overnight?

micahsaul:
@konklone But not to Staging, right?

konklone:
ooh, good point - yes, you should deploy to staging directly

PR #754, opened 2018-01-24
Repair detection of SSLyze results where connection couldn't be established

konklone:
The hacky workaround I was using in #748 and #751 to detect situations when connectivity was not established during SSLyze scanning stopped working. 

The better solution would be a field whose value was just true/false on whether connectivity was established, but in its absence, this fixes the hacky workaround so that affected agencies will see some relief ASAP.


egyptiankarim:
Thanks for pushing this!

PR #761, opened 2018-02-25
Update a11y docs to reflect changed directories and inspect->pshtt

konklone:
The PR to domain-scan at https://github.com/18F/domain-scan/pull/195 moves `scripts/a11y/` to `utils/a11y/`. In anticipation of that change, I've updated the a11y data update docs to reflect this change.

It also pointed to a now-deleted branch, so I updated the `make-lambda-optional` references (which refer to a much older pa11y-specific Lambda tool, not the domain-scan Lambda integration) to point to `master` instead.

It also referenced the `inspect` scanner, which no longer exists in domain-scan, and it looks like the `pshtt` scanner should fill the function identically for what this process uses it for, so I replaced `inspect` with `pshtt`.

cc @micahsaul @gbinal


gbinal:
Looks good.    Thanks!

gbinal:
I’m 👍 to merge.

PR #770, opened 2018-04-11
Set pulse environment to a non-default value.

jmcarp:
Always serve application with waitress and run with debug mode off on
cloud foundry.

Looks like this was the issue affecting @jeremy-gillikin's fork.


konklone:
Interesting! Thanks.

PR #776, opened 2018-04-23
Add logger module

buckley-w-david:
This PR introduces a very basic logging module that CDS has used on another project. Actually using this module is in a later, separate, PR.

Fairly simple addition, a `get_logger` function that just wraps `logging. getLogger` with some formatting, and a function to extract exception messages from chained exceptions.

### DIFF
--- a/data/logger.py
+++ b/data/logger.py
@@ -1,0 +1,25 @@
+import logging
+
+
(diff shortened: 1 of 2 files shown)


konklone:
:+1: It'd be great to see a screenshot or paste of the output in a future PR that makes use of it.

PR #777, opened 2018-04-23
Testing configuration files

buckley-w-david:
This PR adds some sane defaults for linting and type checking configuration.

Currently `pylint` and `mypy` aren't even being used, but we will be integrating it into our CI setup.

These changes combined with the CI configuration are probably the most "CDS specific" of the changes so far, so if you don't want to merge them in, makes sense to me.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,5 +26,6 @@ __pycache__/
 pip-log.txt
 pip-delete-this-directory.txt
 pip-selfcheck.json
+.coverage
 
 todo.txt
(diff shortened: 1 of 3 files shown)


konklone:
Are the effects documented or described anywhere? I'm on board with stronger linting, as long as it's not a straitjacket, but I'd like to know what I'm pulling in.

buckley-w-david:
I suppose I didn't really answer all of your question, in that you'd like to know more about what this linting is.

I could point you to the pylint and mypy documentation, but it's pretty dense and I'm not sure how useful it'd be, is there anything specific you'd like to know?

konklone:
Yeah, I guess I meant if there was anything particularly aggressive or non-standard in the linting settings. But this is fine, I'm up for pulling it back as needed.

PR #796, opened 2018-06-27
change dict_row pattern to instead make use of DictReader

buckley-w-david:
This PR refactors the existing pattern of pulling the first line of a csv into a headers list, then using that list to construct a dict from every row to just using the existing functionality in the python csv library DictReader.

One thing that should be noted is that the behavior of this code will be slightly different in the case of csvs without a header row, but I believe that these would also fail with the existing code, just in another way.


konklone:
Verified output is consistent. Thanks for the cleanup, @buckley-w-david!

PR #797, opened 2018-06-27
Tweak use of setdefault to not collide with existing local variable

konklone:
This tweaks the work from the PR @buckley-w-david filed in #795 to avoid colliding with the existing `subdomains` variable. The `setdefault` code works fine, but the use of a local variable named `subdomains` created problems, as it was already in use in that block.

Without this tweak, the `data.processing` module crashes during data import.


buckley-w-david:
Whoops, good catch.

I thought I tried it out, guess I missed it

